PR #4, opened 2016-08-16
Reference typings/main.d.ts file when compiling

gashcrumb:


kahboom:
+1 tested on OSX and looks good

PR #5, opened 2016-08-16
Update typings devDependency and update readme accordingly

gashcrumb:


kahboom:
+1 Tested on OSX, looks good.

PR #7, opened 2016-09-09
Replace master branch with dev branch completely

kahboom:
## Overview
This is a big PR due to a major deviation from the original `master` branch. This PR is intended to replace the `master` branch with a completely new branch and code base. For reference, here is the background:

>We were using a [starter pack](https://github.com/AngularClass/angular2-webpack-starter) with Angular 2 and Webpack 1 that was used to kick off the [Developer Portal](https://github.com/kahboom/apiman/tree/dev-portal) for apiman and, as such, contained elements of the Developer Portal. We chose this instead of using the starter directly since those elements were beneficial, such as a more developed user interface and a "leaner" project than the original starter pack, since we removed anything we didn't need. However, the Developer Portal was being worked on early in the year and came to a complete halt early on.

>Since Angular 2 was still on the bleeding edge and in alpha (Webpack 2 had not even been released yet), many changes have since been made that also caused the starter pack to change quite a bit in that time period. For instance, structural components were changed in the transition between Webpack 1.x and 2. Changes were expected, but there was a major deviation that would have resulted in rebase/merge hell. Therefore, it was decided that the project was still nascent enough to completely scrap the old Developer Portal skeleton and use the most recent, up-to-date version of the starter pack, which now uses Angular 2 (release candidate 6) and Webpack 2, and then pull in the changes we liked/needed from the Developer Portal, instead of the other way around. To prevent this from happening again in the future, changes to the starter will be pulled in on a more regular basis.

## Changes:
- Merge in the most recent version of the [angular2-webpack-starter](https://github.com/AngularClass/angular2-webpack-starter); resolve conflicts; configure changes.
- Pull features wanted from the Developer Portal into the repository (e.g. referencing templates and stylesheets by URL instead of inline, using SCSS instead of CSS).
- Remove unnecessary assets and features (clean up the starter).
- Set up initial dashboard component and relative assets.
- Add documentation for iPaaS.
- Add necessary third-party dependencies using Webpack's loader and plugin system, including PatternFly and its assets.

## Screenshots
It isn't as pretty as we'd like it to be just yet, as we are running into some issues with using PatternFly's CSS without the use of a CDN and are working with the PF team to figure out the best solution. There is a lot of UI work to be done, as is apparent.

--

cc @gashcrumb


gashcrumb:
Wow, the 'npm install' step is taking a *really* long time...  I may have to continue looking at this on Monday.

gashcrumb:
Yay, it works here!  I think it took about a half hour for `npm install`, but from what I gather that's kinda the norm nowadays with webpack stuff unfortunately.  Doing `npm install --cache-min=999999 -ddd` at least gave me something interesting to look at.  In any case, I'm going to go ahead and merge this now.

kahboom:
@gashcrumb - Not happy about the fact that it took you half an hour to run `npm install` - that definitely shouldn't be the case. I think it took me a few minutes max. I'll clone a clean copy and test it out again. Were you running it on Windows? Over the VPN? What versions of Node.js and NPM?

gashcrumb:
Yep, on Windows, it may have had to build some stuff?  Using node 6.5 and
npm 3.4.1.  It was start and go get lunch long to install :-)

PR #50, opened 2016-10-27
build and run console as a docker container

rawlingsj:


config/webpack.dev.js
```diff
@@ -16,7 +16,7 @@ const NamedModulesPlugin = require('webpack/lib/NamedModulesPlugin');
  * Webpack Constants
  */
 const ENV = process.env.ENV = process.env.NODE_ENV = 'development';
-const HOST = process.env.HOST || 'localhost';
+const HOST = process.env.HOST || '0.0.0.0';
```
gashcrumb: Would it make more sense to just set 'HOST' in the dockerfile rather than changing this default?

rawlingsj: Sure thing.  Done.

PR #79, opened 2016-12-17
Remove moduleId from components

jimmidyson:
This was breaking `npm run build:prod` with error:

    Uncaught TypeError: t.match is not a function

Same error at AngularClass/angular2-webpack-starter#958 with solution at https://github.com/AngularClass/angular2-webpack-starter/issues/958#issuecomment-249502528.

This doesn't seem to cause any problems with dev or production mode, but there may be consequences I'm totally unaware of as an ng2 & webpack n00b. Another support of removing this is in http://stackoverflow.com/questions/37178192/angular2-what-is-the-meanings-of-module-id-in-component#37178813 (lots of upvotes & accepted answer btw) which says:

    If you are using webpack for bundling then you don't need  module.id in decorator. Webpack plugins auto handle (add it)  module.id in final bundle

### DIFF
--- a/src/app/+connections/create/create.component.ts
+++ b/src/app/+connections/create/create.component.ts
@@ -17,7 +17,6 @@ let log = Logger.get('+connections/create');
 const STATE_KEY = 'connection-create-state';
 
 @Component({
-  moduleId: module.id,
   selector: 'connections-create',
   styles: [ require('./create.scss') ],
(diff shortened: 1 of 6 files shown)


jimmidyson:
Again merging so that I can try out the deployment stuff. Doesn't seem to cause any problems.

PR #84, opened 2016-12-20
Load config from server to be able to change API server URL

jimmidyson:


config/webpack.dev.js
```diff
@@ -26,7 +28,7 @@ const METADATA = webpackMerge(commonConfig.metadata, {
   HMR: HMR
 });
 
-const frontendConfig = commonConfig.frontendConfig;
+const frontendConfig = _.merge(commonConfig.frontendConfig, { apiEndpoint: 'http://localhost:8080/v1' });
```
kahboom: So you did end up deciding that the Webpack configuration was the best way to use env-specific settings? I don't mind moving all globals from Angular to Webpack if you'd prefer. I believe it was set up that way before.

config/webpack.common.js
```diff
@@ -241,17 +243,17 @@ module.exports = {
      */
     new CopyWebpackPlugin([
       {
-        from: {glob: './node_modules/patternfly/dist/img/*.*'},
+        from: { glob: './node_modules/patternfly/dist/img/*.*' },
```
kahboom: Thanks for the formatting 👍

jimmidyson: That's from awesome vscode, not intentional :-b

config/webpack.dev.js
jimmidyson: `Globals` could be used instead I guess, More of a `Defaults` than `Globals` perhaps.

jimmidyson:
Turns out this didn't work at all... Looks like the app initializer stuff is a bit broken here, think we need to separate out provider for app config from app state (which is really a service).

PR #85, opened 2016-12-21
Fix dodgy config merge to allow overriding of API server address from loaded config

jimmidyson:


jimmidyson:
Going to merge this to see if it fixes up staging environment.

PR #97, opened 2016-12-22
Aesthetic Changes, Undo State Management & Updates to Detail View

kahboom:
# Overview
Not a ton of functionality in this PR but some key things for the demo.

- Aesthetic changes to reflect designs better across Connections user flow
- Update connection details to pull proper data (ie added breadcrumbs)
- Removed state management for now (buggy)
- Fixed minor bug where check for value is necessary before calling a split method
- Updated Connection model to reflect actual data
- Lots of cleanup, better logging

### Aesthetic Changes

screenshot 2016-12-22 12 51 07

### Disable State Management
This allows the Create Connection work flow to work again.

screenshot 2016-12-22 12 51 21

### Connection Details w/ API Data & Aesthetic Improvements

screenshot 2016-12-22 12 59 10

cc @jimmidyson

### DIFF
--- a/src/app/+connections/component.service.ts
+++ b/src/app/+connections/component.service.ts
@@ -4,8 +4,7 @@ import { Http, Headers, RequestOptions, Response } from '@angular/http';
 // Here were are mostly using observables instead of promises
 import { Observable } from 'rxjs/Observable';
 
-//import * as _ from 'lodash';
-
+import { Globals } from '../app.config';
 import { IComponent } from './component.model';
 //import { IComponentService } from './component.service.interface';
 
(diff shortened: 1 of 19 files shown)


src/app/+connections/component.service.ts
```diff
@@ -18,11 +17,7 @@ export class ComponentService {
 
   errorMessage: string;
   private allComponents: IComponent[];
-
-  //baseUrl: string;
-  //private componentsUrl = 'app/+components/component.data.json'; // URL to JSON file
-  //private componentsUrl = 'http://localhost:9090';
-  private baseUrl = 'http://localhost:9090/v1';
+  private baseUrl = Globals.apiEndpoint;
```
jimmidyson: This should come from injected appState like https://github.com/redhat-ipaas/ipaas-client/pull/97/files#diff-5a78858d0f5523c0c33c96923d35ac24R25

jimmidyson:
As for all the commits, would prefer them to be squashed into functional changes (not necessarily a single commit per PR though) but as we're pressed for time, let's not worry about that for now, discuss in the new year.

PR #101, opened 2017-01-10
Add vscode launch config for easy debugging

jimmidyson:


jimmidyson:
Loving vscode! It rocks for typescript & go, not so much for Java but only because of formatting really.

rhuss:
I tried vscode for go, too. Works nicely but then found out that Gogland (the new IDEA IDE) works even better an already includes some refactorings (extract method etc.) which I use very often. Also, I could already import my keybinding which is big win for an old man like me ;-)

jimmidyson:
Ooo might try that out. The days of a different IDE per language are back again :)

kahboom:
@rhuss Seriously? You make yourself sound like a dinosaur.

rhuss:
@kahboom I am :)

But more seriously I have a very custom set of keybindings (with an emacs-ian origin, you see I'm really from the last ice age :) and I try to configure each tool to fit to these keybindings. Might have been easier to re-memorize stuff for each app, but thats now the way it is ;)

PR #102, opened 2017-01-10
Add mentionbot config

jimmidyson:


jimmidyson:
https://github.com/facebook/mention-bot - really just want to use it to auto assign PRs to someone so we can try to ensure timely reviews.

PR #105, opened 2017-01-10
Move TemplatesListComponent back to TemplatesModule and import from DashboardModule

jimmidyson:
As we discussed yesterday when running through the new architecture.


jimmidyson:
Ug please don't merge. This messes up dashboard route because templates module also has similar child routes & overrides them.

What do you think about a separate routes module per functional area? One module specifically for routing, one for components, etc? Routing module would import the component module obviously.

jimmidyson:
Back ready for review again, introduced templates routes module to work around the problem. Only needed when we have to do something similar for other modules, otherwise we don't need a separate routing module.

gashcrumb:
Seems like sharing components between functional areas is gonna be common, no?  Just wondering if a separate route module should be the standard going forward...

src/app/templates/templates-routes.module.ts
```diff
@@ -0,0 +1,17 @@
...
+  { path: '', component: TemplatesListPage, pathMatch: 'full' },
+];
+
+@NgModule({
+  imports: [
+    RouterModule.forChild(routes),
+    TemplatesModule,
+  ],
+})
+export class TemplateRoutesModule { }
```
gashcrumb: I'm not a fan of the filename but I'm struggling to come up with an alternative :-)

It feels redundant, and it almost feels like it should be `template.routes.ts` but then I get that since it contains an angular module it'd be good to reflect that in the name.

kahboom: +1 @gashcrumb I prefer `template.routes.ts` or `templateRoutes.module.ts`.. but then again I'm a stickler for camelCase over snake-case. If there will only be template routes we should stick with `routes.module.ts`.

gashcrumb: I will say that having 'templates' in the file name helps me out a little only because my most common tool to switch between files doesn't default to showing the path :-)

And +1 camel case filename.  And I suspect we'll hit this with integrations too.

jimmidyson:
> a separate route module should be the standard going forward

I'm kinda resisting that until it definitely is required. With lazy loading it has a teeny impact on loading times but that's probably not worth considering. Let's see how it goes for now - will probably end up with it as you've said. Also we should look at moving more common components (e.g. list) into common module so this kind of functional module sharing won't be required as much I hope.

> I think s/templates-routes.module.ts/templates.routes.ts/ or templatesRoutes.module.ts. Other than that looks good!

From the [Angular 2 style guide](https://angular.io/styleguide#!#02-02) (which I think is a pretty good idea to stick to as much as possible as it gives us a predefined set of guidelines to follow rather than needing to come up with our own - yes I'm lazy):

    Do use dashes to separate words in the descriptive name.

and

    Do use conventional type names including .service, .component, .pipe, .module, .directive. Invent additional type names if you must but take care not to create too many.

So I'd prefer to stick with `templates-routes.module.ts`, even with singular `template-routes.module.ts`. WDYT?

gashcrumb:
Well that's sound enough reasoning for me :-)

kahboom:
GAH! ;D

jimmidyson:
I normally merge on one approved review from core maintainer, but I shouldn't self-merge either :-b Naughty Jimmi.

kahboom:
@jimmidyson That sounds reasonable to me! Nothing wrong with self-merging once it's been approved from at least one core maintainer imo.

PR #107, opened 2017-01-11
Ensure loading spinner is in consistent position across lists

jimmidyson:
Because integrations list is a slightly different view (why btw)? The loader wasn't consistent with other list views.

It's subtle, but it's there - you have to look closely in these gifs to see it, have to figure out a better way to highlight changes like this without being too onerous.

Here's a before:

before

And an after:

after

### DIFF
--- a/src/app/common/loading/loading.component.html
+++ b/src/app/common/loading/loading.component.html
@@ -1,4 +1,4 @@
-<div *ngIf="loading">
+<div class="loading" *ngIf="loading">
   <div class="spinner spinner-lg"></div>
 </div>
 <ng-content *ngIf="!loading"></ng-content>
--- a/src/app/common/loading/loading.component.scss
+++ b/src/app/common/loading/loading.component.scss
@@ -1,0 +1,3 @@
+.loading {
+  padding-top: 20px;
+}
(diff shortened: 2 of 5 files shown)


gashcrumb:
Sorry, totally should have approved this earlier :-)

kahboom:
@jimmidyson @gashcrumb - We should really take a minute to create issues for things like this, minor as they are. It helps keep track of the things we've accomplished on a day-to-day basis, and they don't have to be huge accomplishments. I am happy to volunteer to create issues for things you are working on but don't have time to create an issue for- just shoot me a quick message on IRC. It honestly doesn't bother me and I think helps keep everyone on the same page. :)

jimmidyson:
We can have a bot that validates if an issue exists for it perhaps?

kahboom:
@jimmidyson - LOL you crack me up with the bots. I think that's a good idea. Btw, am I the only one using the Kanban board and GitHub issues? :( https://github.com/redhat-ipaas/ipaas-client/projects/6 It's okay as I mostly use it for myself, but when I see you guys are making progress on something I usually just move it around for you lol (#ocd). I can use it just for my issues if anything.

jimmidyson:
Bots ftw... I want to use bots to enable & enforce the processes we want in place.

PR #111, opened 2017-01-13
lets add a spike at a routing doc

jstrachan:


jimmidyson:
As this stuff is going to be common to all our angular2 development, I'd like to put this into a separate docs-only repo so it can be easily referenced, updated, published, etc. WDYT?

@kahboom Is this something you'd like to look after for us?

jstrachan:
@jimmidyson sure, sounds good to me - lets make sure we clearly link to it from the contributing/readme files so folks don't lose it. Could also go into naming conventions for the various view/view-page/view-toolbar kinda stuff too maybe?

jimmidyson:
> Could also go into naming conventions for the various view/view-page/view-toolbar kinda stuff too maybe?

Yeah we need to do that. For now, it's enough to just reference the Angular 2 style guide at https://angular.io/styleguide, but we can copy over what we need as we need it.

jimmidyson:
I've created https://github.com/redhat-ipaas/ipaas-ui-style-guide using asciidoctor for documentation. Published on push to master branch to https://redhat-ipaas.github.io/ipaas-ui-style-guide/

PR #112, opened 2017-01-13
Add create and update to AbstractStore

jimmidyson:


jimmidyson:
Ping @jstrachan

jimmidyson:
Once we have a nice error component then we should have an errors service that are updated from stuff like this.

jstrachan:
lgtm [merge]

kahboom:
@jimmidyson - Do you think there is a use case here for Delete in AbstractStore?

jimmidyson:
Yep. Can you add it in please?

kahboom:
Thanks, I figured. I'll create a separate issue for it.

PR #115, opened 2017-01-17
Upgrade dependencies

jimmidyson:


kahboom:
Would still like to see exact patch semantic versioning but other than that, looks good to me.

package.json
```diff
@@ -5,65 +5,61 @@
...
-    "zone.js": "^0.7.4"
+    "zone.js": "^0.7.5"
   },
   "devDependencies": {
-    "@angular/compiler-cli": "^2.4.1",
-    "@types/jasmine": "^2.5.38",
-    "@types/lodash": "^4.14.44",
-    "@types/node": "^6.0.52",
+    "@angular/compiler-cli": "2.4.3",
+    "@types/jasmine": "^2.5.40",
```
jimmidyson: Now that we use yarn I don't think we need to bother that. Upgrading dependencies should be a considered action using `yarn upgrade ...` which will update `yarn.lock` accordingly. Normal dev should just be `yarn` (or `yarn install`) to install dependencies as defined in lock file.

src/app/app.component.spec.ts
```diff
@@ -17,14 +17,14 @@ describe('AppComponent', () => {
   });
 
   it('should create the app', async(() => {
-    let fixture = TestBed.createComponent(AppComponent);
-    let app = fixture.debugElement.componentInstance;
+    const fixture = TestBed.createComponent(AppComponent);
```
jimmidyson: `const` ftw. `yarn lint:fix` will do a lot of that for you btw.

jimmidyson:
@kahboom Have you tried locally? I've seen intermittent huge slowdown in builds with these upgrades & not sure if it's something local for me or a more general problem.

kahboom:
@jimmidyson - A check failed, do you still want me to merge it in? Per Jenkins it looks like we need to set a CHROME_BIN env var for Karma to run: `No binary for Chrome browser on your platform.
  Please, set "CHROME_BIN" env variable.`

Also, typical errors with `fsevents` since we are running on Linux, and `fsevents` should only be run on OS X or macOS, but npm can be stupid like that:

```
npm WARN optional SKIPPING OPTIONAL DEPENDENCY: fsevents@^1.0.0 (node_modules/angular-cli/node_modules/chokidar/node_modules/fsevents):
npm WARN notsup SKIPPING OPTIONAL DEPENDENCY: Unsupported platform for fsevents@1.0.17: wanted {"os":"darwin","arch":"any"} (current: {"os":"linux","arch":"x64"})
```

I think there is a way to circumvent that, but not worth it considering it's just a warning and not the actual cause of the build failure.

kahboom:
> @kahboom
>Have you tried locally? I've seen intermittent huge slowdown in builds with these upgrades & not sure if it's something local for me or a more general problem.

I have not. I'll give it a whirl now.

kahboom:
I'm nitpicking and this is off topic (personal note, not worth a GH issue, but for my next PR), but we need to update the `README` to have `npm start` as the start instructions. No way I'm installing anything globally on my machine; we need to be using the npm binaries (as we are), but avoid users thinking that they need to install angular-cli globally.

PR #135, opened 2017-02-01
Update angular cli

gashcrumb:
`latest` doesn't seem to matter with the yarn.lock file, does it?


gashcrumb:
Interesting, they'll be changing the package name to `@angular/cli`, so something to watch out for I guess

jimmidyson:
Ooo nice warning :)

jimmidyson:
[merge]

PR #138, opened 2017-02-02
Patternfly update, replace ng-bootstrap with ng2-bootstrap

gashcrumb:
Also fixed that pesky stating URL since @jimmidyson changed it yet again ;-)

Included the bootstrap css just in case, though I think the patternfly css *should* cover any styles used in `ng2-bootstrap`.  But you never know.

Initialized a few modules from `ng2-bootstrap` that we'll definitely be needing but there's more, full demo here -> https://valor-software.com/ng2-bootstrap/#/


gashcrumb:
Oh, fixes https://github.com/redhat-ipaas/ipaas-ui/issues/136

kahboom:
As much as I wanted to use Bootstrap 4 and ng-bootstrap, it turns out that ng2-bootstrap really was the better option. Also @gashcrumb , you've rendered my only [open PR](https://github.com/redhat-ipaas/ipaas-ui/pull/137) useless. ;) Trying to stick to tiny PRs as I have a tendency to make huge ones.

Issue #120, opened 2017-01-17
Reusable Toolbar Component


kahboom:
Create a component for the toolbar that allows it to be reused across multiple components, likely as a child component. May be related to #116 or #117.

Issue #116, opened 2017-01-17
Shared Feature Module


kahboom:
Create a Shared Feature module, as specified in the Angular 2 official style guide [here](https://angular.io/styleguide#04-10). We would be using this for things that are not singletons, and will be instantiated for each component that uses it, such as filter pipes. Possible challenges are getting it to work properly with AOT and lazy loading (and angular-cli), but may not run into any issues at all.

For non-singletons we will be using a Shared Feature Module (see #117).

gashcrumb:
We have the `common` module, think this is all set.

Issue #89, opened 2016-12-22
Reusable Filter w/ Support for Multiple Properties


kahboom:
Add filtering using multiple properties in Angular 2. This feature should be reusable and, as such, a part of the Shared Feature Module (see #116).

## Reference
- [JIRA User Story](https://issues.jboss.org/browse/IPAAS-145)
- [Connections: Connection Library](https://github.com/hawtio/hawtio-ipaas/issues/42)

### InVision Designs:
- [Connection Library](https://redhat.invisionapp.com/share/ZQ9OF7V5Y)

kahboom:
This will be part of the Shared Feature Module (see #116). Sorting is already supported; filtering still needs to be completed, so I will update this issue to reflect that.

gashcrumb:
This is halfway done, the filters there but it only supports filtering on one property.

gashcrumb:
We kinda get this now via the patternfly toolbar component.

PR #142, opened 2017-02-07
re #89 #120 Filter text service, swap parent+child selectors, cleanup

kahboom:
This PR is for part of the work required for #89 (Reusable Filter w/ Support for Multiple Properties) and #120 (Reusable Toolbar Component). Reusability of the filter and toolbar may or may not be possible for this sprint, as that would be considered refactoring and an enhancement, but the actual filter and toolbar for each component will be provided as a requirement. The UI doesn't get updated with the newly filtered list, but that's okay (okay maybe not, but it will have to wait for the next PR).

There are some `console.log()`s in the code that are there due to the fact that this PR does not fully implement the toolbar and filter for multiple properties, but it resolves an inherent problem that was causing issues when sharing data across parent and sibling components. This should also facilitate adding in any type of storage, as previously discussed, once that is underway.

Changes:
- In Connections component, include child directives in parent directives, not the other way around.
- Add `FilterTextComponent` and `FilterTextService` in `IPaaSCommonModule`, which will be the Shared Feature module in the future (see #116 ).
- Update Zone.js to latest version, which resolves an issue with the stack trace being pretty useless (see [here](https://github.com/angular/zone.js/issues/427)).
- Add instructions for running the linter in `README.md`.
- Prevent calling of `UserService` directly from the DOM; moved to the component logic and created a local variable.
- Fix tests that failed after swapping parent/child directives. This also required making some properties public (nothing unsafe).
- Lock dependency versions in `package.json` for good measure and best practices.
- Update some TSLint rules that were driving me insane (namely, requiring a space before each comment, as well as requiring every declared local to be used).
- Remove some buttons from the UI that are not applicable to this sprint (e.g. Create a Connection, Create a Template).

In Progress (next PRs):
- Update UI with actual filtered lists (Angular 2 making my life difficult with filters or "pipes").
- UI implementations of designs for Integration List + Detail (if they are not already completed).
- Update other components with proper parent/child component structure.
- Refactor to make services and components reusable.
- Set up and add E2E tests, including with CD/CI.
- Improve unit tests and add where needed.

I also don't like the way the filter works at the moment. Besides the obvious need for it to update the UI, there is a transient error when typing too quickly and, more importantly, there are several `EventEmitter`s to keep track of, which terrify me of memory leaks in the future. I'm going to improve on this soon, but it works for now and I need to keep going. If you have any suggestions on how to improve this quickly, feel free to let me know.


src/app/common/filter-text/filter-text.service.ts
```diff
@@ -0,0 +1,44 @@
...
+
+    if (data && props && originalList) {
+      data = data.toLowerCase();
+
+      originalList.filter(item => {
+        let match = false;
+
+        for (const prop of props) {
+
+          console.log('prop: ' + JSON.stringify(prop));
```
gashcrumb: Be good to take out these console.log statements, or we should perhaps use a proper logging subsystem so we don't clutter up the dev console

src/app/common/filter-text/filter-text.service.ts
```diff
@@ -0,0 +1,44 @@
...
+
+          if (item[prop].toString().toLowerCase().indexOf(data) > -1) {
+            console.log('match found: ' + JSON.stringify(item[prop]));
+            match = true;
+
+            filteredList.push(item);
+
+            console.log('match: ' + JSON.stringify(match));
+            console.log('item: ' + JSON.stringify(item));
```
gashcrumb: Same as above.

src/app/common/filter-text/filter-text.service.ts
```diff
@@ -0,0 +1,44 @@
...
+            break;
+          }
+
+          console.log('filteredList: ' + JSON.stringify(filteredList));
```
gashcrumb: nuther one, also for really long lists this will cause a lot of text to go to the dev console, better to not use JSON.stringify() and let the browser render the array, like:

```
console.log('filteredList: ', filteredList);
```

src/app/connections/list/list.component.html
```diff
@@ -1,47 +1,8 @@
...
-            <p>{{ connection.description | truncate : truncateLimit : truncateTrail }}</p>
-          </div>
-        </div>
-      </div>
-    </div>
+  <div class='connections list'>
+    <ipaas-connections-list-toolbar [childConnections]="connections | async"
+                                    (change)="updateConnections($event)"></ipaas-connections-list-toolbar>
+    <ipaas-connections-list-page [connectionList]="connections"
+                                 [loading]="loading"></ipaas-connections-list-page>
```
gashcrumb: This is effectively swapping the list-page and the list components, but I don't understand why.  The way I understood the previous approach is the page handles loading the data from the store, and passes it along to the list/toolbar views to render/filter appropriately.  Now it looks like the `page` component is loading and rendering

Note, I have a vested interest in this as I'm re-using the `ipaas-connections-list` and `ipaas-connections-list-toolbar` components in the integration editor.  :-)

src/app/connections/list/list.component.ts
```diff
@@ -1,20 +1,34 @@
...
+
+  updateConnections(event) {
+    console.log('Event from updateConnections: ' + JSON.stringify(event));
+    this.connections = event;
+  }
 
-  @Input() loading: boolean;
+  ngOnInit() {
+    this.store.loadAll();
+  }
```
gashcrumb: See above, I'm not sure I understand why this code is moving from the page to the list.

src/app/common/filter-text/filter-text.component.ts
```diff
@@ -0,0 +1,25 @@
...
+
+  constructor() {
+    this.changed = new EventEmitter<string>();
+  }
+
+  clear() {
+    this.filter = '';
+  }
+
+  filterChanged(event: any) {
```
jimmidyson: We should be debouncing these events. How about using rxjs something like http://plnkr.co/edit/98ijjY?p=preview (see script.ts):

```typescript
const input$ = Rx.Observable
  .fromEvent(document.getElementById('hello'), 'keyup')
  .map(x => x.currentTarget.value)
  .debounceTime(500);
  
const updated$ = document.getElementById('updated');
input$.subscribe(x => updated.innerHTML = x);
```

src/app/common/filter-text/filter-text.service.ts
jimmidyson: Yeah but please in another PR, this is already big enough.

src/app/common/filter-text/filter-text.service.ts
jimmidyson: Right, we should be logging objects when we need to for much easier inspection.

src/app/connections/connections.module.ts
```diff
@@ -4,17 +4,17 @@ import { FormsModule } from '@angular/forms';
...
-import { ConnectionViewPage } from './view-page/view-page.component';
+import { ConnectionViewPageComponent } from './view-page/view-page.component';
 import { ConnectionViewWrapperComponent } from './view-wrapper/view-wrapper.component';
 import { ConnectionViewToolbarComponent } from './view-toolbar/view-toolbar.component';
 import { ConnectionViewComponent } from './view/view.component';
 
 const routes: Routes = [
-  { path: '', component: ConnectionsListPage, pathMatch: 'full' },
-  { path: ':id', component: ConnectionViewPage, pathMatch: 'full' },
+  { path: '', component: ConnectionsListComponent, pathMatch: 'full' },
```
jimmidyson: The whole point in using Page suffix is to indicate that it's a wrapper page, a top level component in this case. I don't really like the idea of changing that personally.

src/app/connections/list-page/list-page.component.spec.ts
```diff
@@ -1,39 +1,32 @@
...
-import { ConnectionsListPage } from './list-page.component';
 import { ConnectionsListComponent } from '../list/list.component';
+import { ConnectionsListPageComponent } from './list-page.component';
 import { ConnectionsListToolbarComponent } from '../list-toolbar/list-toolbar.component';
-import { StoreModule } from '../../store/store.module';
 
-describe('ConnectionListPage', () => {
-  let component: ConnectionsListPage;
-  let fixture: ComponentFixture<ConnectionsListPage>;
+describe('ConnectionsListComponent', () => {
```
jimmidyson: This would have the wrong output description now that it's been renamed, but I'd prefer not to rename it at all as I said previously.

src/app/connections/list/list.component.html
jimmidyson: Right - this is fundamental to the container/presentational components idea. The page is what knows what a data to load, the list just displays it. That means we can reuse the list component as @gashcrumb said above.

src/app/common/filter-text/filter-text.service.ts
kahboom: Not a problem, I completely agree. I'll remove them. Btw, we used to have that logger from hawtio that @gashcrumb added in, not sure where it went. Would be nice to have env-specific logging for when forgetful devs like me forget to remove logs.

src/app/connections/list/list.component.html
kahboom: Maybe this is because I haven't really looked at the integration editor designs, or maybe because I'm still thinking like Angular 1.x, but I'm confused about the previous approach. 1) Why would we be re-using lists if we are already going to be using a data store and a service to fetch the data? How many views will be using this same list to warrant it being its own component? Normally just a parent component for the Connections model, and child `list` and `detail` components suffice.* 2) The reason for this change is the idea that the parent is `list.component.ts` would be the only place that fetches the data and has several directives it passes that data to. In this case we can pass them whatever model we want. Looking at the way we organized files there was never any indication to me that the page was responsible for loading the data from the store and that it was the primary container. For that we can create a separate, reusable, agnostic, data-fetching, list-generating component that accepts any model and returns a list.

* The only exception to this are reusable components, like toolbars, that would be children components, and should definitely inherit from parent components that pass them whatever data you indicate it should have access to.

src/app/connections/list/list.component.ts
kahboom: Because any directive it contains can inherit the data more easily. See [here](https://angular.io/docs/ts/latest/cookbook/component-communication.html).

src/app/common/filter-text/filter-text.component.ts
kahboom: Sure, sounds good.

src/app/connections/connections.module.ts
kahboom: Ah I see, I thought we were ending all components with `Component`. We should add all of this to the style guide.

src/app/common/filter-text/filter-text.service.ts
gashcrumb: Yeah, lemme get a PR in soon to add that in, it was `js-logger`, might see if there's one that has typescript defs.

src/app/connections/list/list.component.html
gashcrumb: So in the integration editor we've two pages that effectively re-use the connection list + toolbar that the user will use to select the start and end connection.  Wait, should I be saying `iPAAS Connector` here?

In any case, I only want to use the list component and the toolbar component, and I want my wrapper page to load the data that those two components look at.  I also am planning on adding an `@Output` to that list component so that I can let the user click on a connection to select it.

src/app/common/filter-text/filter-text.service.ts
gashcrumb: @jimmidyson you're not gonna like mine then :-)

PR #143, opened 2017-02-08
Initial work for creating/editing integrations

gashcrumb:
Most work has been figuring out the routing and re-using components from connections to implement the designs.  No styling work has been done, so excuse the look of the screenshot(s), I kinda sorta started using the patternfly classes for the wizard pattern, but the design doesn't really match that pattern 100%.  Anyways, this was more about just getting elements on the page, not necessarily where they need to be :-)

And the difference between the two is subtle, look at the URL and also the `name` field at the top-left

via `Create Integration` link:

via-create-link

via clicking the `Edit` link from the list of integrations:

via-edit-link

Other changes (yes, there's lots of 'em)

* refactored out routing in the connections module to use a separate routing module.
* create a 'flow-view' component
* Add wrapper pages in the integration module for selecting and showing connections
* Change the stores so that they only provide the returned data, not a restangularized object, as updates to objects should always be via the relevant data store.

I'll apologize in advance for the merge commits, I had forgotten to configure my git locally to rebase by default.

TODO
* will probably rename `create-page` to something else
* will likely move directories under `create-page` around
* wire stuff up

### DIFF
--- a/src/app/integrations/create-page/flow-view/flow-view.component.ts
+++ b/src/app/integrations/create-page/flow-view/flow-view.component.ts
@@ -1,0 +1,22 @@
+import { Component, Input, OnInit } from '@angular/core';
+import { Observable } from 'rxjs/Observable';
+
+import { Integration } from '../../../store/integration/integration.model';
+
+@Component({
+  selector: 'ipaas-integrations-flow-view',
+  templateUrl: './flow-view.component.html',
+  styleUrls: ['./flow-view.component.scss']
+})
+export class FlowViewComponent implements OnInit {
+
+  @Input() integration: Observable<Integration>;
+  i: Integration;
+
+  ngOnInit() {
+    this.integration.subscribe((integration) => {
+      this.i = integration;
+    });
+  }
+
+}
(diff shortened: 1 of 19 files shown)


gashcrumb:
Working on fixing the tests, so don't merge this yet :-)

gashcrumb:
K, rebased against master, quick hack to fix the broken tests, definitely a TODO item to add tests for some of the new components as they get more fleshed out.

src/app/integrations/create-page/flow-view/flow-view.component.ts
```diff
@@ -0,0 +1,22 @@
...
+  templateUrl: './flow-view.component.html',
+  styleUrls: ['./flow-view.component.scss']
+})
+export class FlowViewComponent implements OnInit {
+
+  @Input() integration: Observable<Integration>;
+  i: Integration;
+
+  ngOnInit() {
+    this.integration.subscribe((integration) => {
```
jimmidyson: Shouldn't need to subscribe here. Can just pass it through via async pipe in template perhaps?

gashcrumb: Ah yeah, I think I hadn't yet arrived at the `getOrCreate()` function in the stores, so this was my way of getting around the fact that we're either creating an empty integration object or working with one that's been fetched.  Also I haven't 100% decided how much this controller is going to handle vs the containing page, so just left this code in place for now.

PR #145, opened 2017-02-09
Rm ref to UserSvc from HTML, update zone.js, confs, update UI for scope of sprint

kahboom:
Changes:
- Rm ref to UserService from HTML, create local ref in logic instead
- Rm ref to create or edit Templates or Connections in HTML
- Rm caret from Create Integration btn which implies more than one way to create
- Update Zone.js to latest version for improved error logging, which resolves an issue with the stack trace being pretty useless (see [here](https://github.com/angular/zone.js/issues/427)).
- Add instructions for running the linter in `README.md`. Rm resource section (will update this in the iPaaS website).
- Lock dependency versions in `package.json` for good measure and best practices, regardless of using Yarn.
- Update some TSLint and `tsconfig.json` rules that were driving me insane (namely, requiring a space before each comment, as well as requiring every declared local to be used).

In Progress (next PRs):
- Filterable, sortable lists should I not tear my hair out before then.
- Remaining UI implementations of designs for Integration List + Detail (if they are not already completed).
- Connections List deleting a connection w/ toast notifications & modal confirmations.
- Integrations "".
- Refactor to make certain services and components reusable (reaching here).
- Set up and add E2E tests, including with CD/CI.
- Improve unit tests and add where needed.


src/app/connections/list/list.component.html
```diff
@@ -5,9 +5,9 @@
         <div class="card-pf card-pf-view card-pf-view-xs">
           <div class="card-pf-body">
 
-            <!-- Card Dropdown -->
-
+            <!-- Card Heading -->
             <div class="card-pf-heading-kebab">
+              <!-- Card Actions --
```
kahboom: Ah that was to comment out the section. Would you like for me to change it to have a separate commented out block from the title?

src/app/integrations/list-toolbar/list-toolbar.component.html
```diff
@@ -45,8 +45,10 @@
           </button>
       </div>
       <div class='toolbar-pf-action-right'>
-        <button type='button' class='btn btn-primary dropdown-toggle' [routerLink]="['create']">Create an Integration&nbsp;<span class='fa fa-angle-down'></span>
-          </button>
+        <button type='button' class='btn btn-primary dropdown-toggle' [routerLink]="['create']">
+          Create an Integration
+          <!--&nbsp;<span class='fa fa-angle-down'></span>-->
```
kahboom: Sure thing.

src/app/connections/list/list.component.html
gashcrumb: Ah totally missed that, funny really only noticed it because github highlighted it so obviously.  Yeah, may as well comment out the block in a separate comment.

Issue #130, opened 2017-01-26
Improve Configurability per Environment


kahboom:
It would be very useful to be able to switch, for example, API and auth URLs from a configuration file without worrying about having to switch it back when you are done testing or coding. This file shouldn't get checked in. There would be an example configuration file that users would rename to `config.json`. It adds an extra step in the setup but provides added convenience.

See my proposed solution here and let me know what you think: https://gist.github.com/kahboom/7d1ab66fa0c85bbf20594c9b1cc477cd

Benefits:
- Improved configurability. It would help us reference constant values that can easily be changed from a top level without being hardcoded into the `app.component.ts` file or anywhere else. Values can also be overridden based on the environment while defaulting to `development`.
- DRYer code. We won't be repeating ourselves if we keep all constant values that are browser-side in a single place. The configuration service would still be very helpful as that is where we could write the default to `development` functionality.
- Better unit tests. It also makes it a million times easier to write unit tests and make sure the tests are actually working as expected (failing when they need to, not just passing), as you can easily switch out the title, for instance, and see if it makes the test fail as it should.
- Better E2E testing. This could also be very useful for E2E testing, where you often want to test both remote and local setups (if the dev decides to run E2E tests manually at any given point they should be able to, not only via CI).

Is it overkill refactoring or unnecessary? Possibly. I think the tradeoffs are worth it, but I would love to hear any thoughts.

jimmidyson:
Can't we use cli [environments](https://github.com/angular/angular-cli#build-targets-and-environment-files) for this?

kahboom:
@jimmidyson - Instead of a string for the `env` key? Yes, that'd be a good option, but specifying the environment isn't the main focus here, it's using configuration settings (via JSON objects) that are specific to them, which would require not checking in the `config.json` file. What do you think?

PR #148, opened 2017-02-10
[WIP] Add authentication via keycloak

jimmidyson:
This also includes necessary separating out of environment config so we can provide
defaults for development and override at runtime by dropping in separate config file.

### DIFF
--- a/src/app/app.component.html
+++ b/src/app/app.component.html
@@ -1,5 +1,5 @@
 <header *ngIf="loggedIn">
-  
+
   <!-- Navbar: Logged Out -->
   <nav class="navbar navbar-fixed-top navbar-light loggedOut" *ngIf="!loggedIn">
     <!-- Logo -->
@@ -7,7 +7,7 @@
       <img src="{{ logoWhiteBg }}">
     </a>
   </nav>
-  
+
   <!-- Navbar: Logged In -->
   <nav class="navbar navbar-pf-vertical">
     <div class="navbar-header">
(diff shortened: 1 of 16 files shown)


jimmidyson:
Should have said please don't merge as this requires manfredsteyer/angular-oauth2-oidc#8 to be merged to support hybrid oidc flow and a release of that module. If it doesn't happen quickly enough i'll release from my fork.

kahboom:
@jimmidyson - Let me know when it's ready to be reviewed.

jimmidyson:
@kahboom It's code complete, just not ready for merge until that module is released so feel free to review whenever.

jludvice:
Hi @jimmidyson +1 for:
>This also includes necessary separating out of environment config so we can provide
defaults for development and override at runtime by dropping in separate config file.

Could you please provide informations on how to setup oAuth for e2e and unit tests ?

kahboom:
@jludvice - 👍  That's one of the main reasons I wanted this feature! See issue #130

jimmidyson:
> Could you please provide informations on how to setup oAuth for e2e and unit tests ?

Right now we just need a user (or two) that can authenticate to the cluster that the e2e tests run against.

src/app/app.component.html
```diff
@@ -87,6 +87,6 @@
     -->
   </ul>
 </div>
-<main class="container-fluid container-pf-nav-pf-vertical">
+<main class="container-fluid container-pf-nav-pf-vertical" *ngIf="loggedIn">
```
kahboom: Not sure this would be a good way to do this in the future, but is fine for now. We can eventually include different templates depending on whether or not the user is logged.

kahboom: Sorry, should have specified, using `*ngIf` from within each template.

src/environments/environment.ts
```diff
@@ -3,6 +3,17 @@
 // `ng build --env=prod` then `environment.prod.ts` will be used instead.
 // The list of which env maps to which file can be found in `angular-cli.json`.
 
-export const environment = {
+export const environment = Object.freeze({
   production: false,
-};
+  config: {
+    apiEndpoint: 'http://localhost:8080/api/v1',
+    title: 'DEVELOPMENT - Red Hat iPaaS',
```
kahboom: Good idea to include this in the title.

src/app/app.component.html
jimmidyson: Yeah I realise that's a bit rubbish and agree we should fix up in future, but works for now.

src/environments/environment.ts
jimmidyson: I normally like to make it more obvious, e.g. banner or something on page but this is again, ok for now.

jimmidyson:
The current flow looks like [this](https://www.youtube.com/watch?v=GwEwt-lLIC8) from deployment of iPaaS, through logging in, to being able to use the UI.

kahboom:
Lord help us

PR #149, opened 2017-02-10
Rework connection list to trigger a selected event so parent component can control what to do

gashcrumb:
This allows the create integration page to re-use the connection list for selecting connections.

Ultimately we'll need to sort having a kebab drop-down within the clickable connection element, an issue we've handled before [this way](https://github.com/redhat-ipaas/ipaas-ui/blob/master-original/src/app/%2Bconnections/library/library.component.ts#L85-L87), though maybe if we use ngClick instead to trigger the drop-down we wouldn't have to resort to that workaround.


jimmidyson:
[merge]

kahboom:
Ty @jimmidyson !

Issue #52, opened 2016-10-27
e2e tests won't run


gashcrumb:
Been looking into this issue this afternoon:

```
~/Work/src/hawtio-ipaas
] npm run e2e

> hawtio-ipaas@1.0.0 pree2e /home/gashcrumb/Work/src/hawtio-ipaas
> npm run webdriver:update -- --standalone


> hawtio-ipaas@1.0.0 webdriver:update /home/gashcrumb/Work/src/hawtio-ipaas
> npm run webdriver-manager update "--standalone"


> hawtio-ipaas@1.0.0 webdriver-manager /home/gashcrumb/Work/src/hawtio-ipaas
> webdriver-manager "update"

selenium standalone is up to date.
chromedriver is up to date.

> hawtio-ipaas@1.0.0 e2e /home/gashcrumb/Work/src/hawtio-ipaas
> npm run protractor


> hawtio-ipaas@1.0.0 protractor /home/gashcrumb/Work/src/hawtio-ipaas
> protractor

[15:24:07] I/direct - Using ChromeDriver directly...
[15:24:07] I/launcher - Running 1 instances of WebDriver

/home/gashcrumb/Work/src/hawtio-ipaas/node_modules/selenium-webdriver/http/index.js:365
      onError(new Error(message));
              ^
Error: ECONNREFUSED connect ECONNREFUSED 127.0.0.1:40954
    at ClientRequest.<anonymous> (/home/gashcrumb/Work/src/hawtio-ipaas/node_modules/selenium-webdriver/http/index.js:365:15)
    at emitOne (events.js:96:13)
    at ClientRequest.emit (events.js:188:7)
    at Socket.socketErrorListener (_http_client.js:310:9)
    at emitOne (events.js:96:13)
    at Socket.emit (events.js:188:7)
    at emitErrorNT (net.js:1276:8)
    at _combinedTickCallback (internal/process/next_tick.js:74:11)
    at process._tickCallback (internal/process/next_tick.js:98:9)
From: Task: WebDriver.createSession()
    at acquireSession (/home/gashcrumb/Work/src/hawtio-ipaas/node_modules/selenium-webdriver/lib/webdriver.js:62:22)
    at Function.createSession (/home/gashcrumb/Work/src/hawtio-ipaas/node_modules/selenium-webdriver/lib/webdriver.js:295:12)
[...]

PR #152, opened 2017-02-13
[relates #52] Fix SpecReporter require in protractor config

jludvice:
According to lathonez/clicker#205 there was api break in SpecReporter 3.

With this change I'm able to run `yarn e2e`, but the tests will fail anyway, because browser is redirected to github for oAuth.

### DIFF
--- a/protractor.conf.js
+++ b/protractor.conf.js
@@ -2,7 +2,7 @@
 // https://github.com/angular/protractor/blob/master/docs/referenceConf.js
 
 /*global jasmine */
-var SpecReporter = require('jasmine-spec-reporter');
+var SpecReporter = require('jasmine-spec-reporter').SpecReporter;
 
 exports.config = {
   allScriptsTimeout: 11000,
--- a/protractor.conf.js
+++ b/protractor.conf.js
@@ -2,7 +2,7 @@
 // https://github.com/angular/protractor/blob/master/docs/referenceConf.js
 
 /*global jasmine */
-var SpecReporter = require('jasmine-spec-reporter').SpecReporter;
+let SpecReporter = require('jasmine-spec-reporter').SpecReporter;
 
 exports.config = {
   allScriptsTimeout: 11000,


protractor.conf.js
```diff
@@ -2,7 +2,7 @@
 // https://github.com/angular/protractor/blob/master/docs/referenceConf.js
 
 /*global jasmine */
-var SpecReporter = require('jasmine-spec-reporter');
+var SpecReporter = require('jasmine-spec-reporter').SpecReporter;
```
jludvice: In  lathonez/clicker#205 they suggest
```javascript
let SpecReporter = require('jasmine-spec-reporter').SpecReporter;
```

But idea complains about `let` unsupported in this version of javascript. `yarn e2e` worked with `let` though.

Which one is correct? `var` or `let`?

jimmidyson: We require es6/es2015 (whatever it's called nowadays) for build time at least, so prefer `let`.

protractor.conf.js
```diff
@@ -2,7 +2,7 @@
 // https://github.com/angular/protractor/blob/master/docs/referenceConf.js
 
 /*global jasmine */
-var SpecReporter = require('jasmine-spec-reporter');
+let SpecReporter = require('jasmine-spec-reporter').SpecReporter;
```
jludvice: @jimmidyson var changed to let

kahboom:
@jimmidyson - This looks fine, but we still have a problem. @jludvice was able to fill out the GH login form, take a screenshot from Protractor, but this required exporting personal credentials in shell, which we obviously can't do.

@jludvice - Hope you don't mind, I'm copying and pasting my email reply to you for @jimmidyson to provide some insight on:

> So, we'd need to create an OAuth application from the GitHub repository and provide those token credentials rather than your own, like this. I think other than that we can also use jimmibot for this since jimmibot is already a user in the organization as well, but the first approach is probably best practice and still closely emulates a user. Hope that helps. I'll take a look at the PR now, apologies for the delay. :)

Let me know what you think, but either way this should probably be addressed in a separate PR/GH issue.

kahboom:
btw @jludvice thanks for referencing the GH issue 👍

jimmidyson:
Until #148 is merged, auth is hard coded against the https://api.rh-idev.openshift.com, the OpenShift cluster we use for dev. Once #148 is merged, this introduces Keycloak so we can be a bit more flexible with auth perhaps. Although we'll still use Keycloak to delegate auth to OpenShift and further e2e tests will have to be properly authenticated on the OpenShift cluster we run tests against.

I wouldn't be happy to give out @jimmibot's credentials because it's an org admin. I would prefer to create an ipaas-test-bot user or something that is a member of the org but has no karma to push, etc. We should never be adding credentials to scripts, but can provide that through the environment via secrets probably comsumed via environment variables.

kahboom:
@jimmidyson - Sorry just saw your last message after reviewing the other PR (https://github.com/redhat-ipaas/ipaas-ui/pull/148). Is it okay to merge then?

Also, I agree about not giving out @jimmibot 's credentials, but why can't we use a registered OAuth app like [here](https://developer.github.com/guides/basics-of-authentication/#registering-your-app)? Does it really have to be a user for any particular reason, or just preference? Either way, it's simple enough to create a bot user, just curious.

jimmidyson:
@kahboom Apps are used for users to grant them access to their resources. The app doesn't log in, the user logs in and grants the app authority to act on their behalf. In our case, the OpenShift cluster is currently using GitHub for auth so has its own registered app which the cluster is configured to use. You still log in as _you_, not the app.

jimmidyson:
Oh and yes I'm happy to merge this, as long as it isn't a blocker for PRs yet, until we've sorted out the auth stuff properly.

kahboom:
@jludvice - Leaving this open until https://github.com/redhat-ipaas/ipaas-ui/pull/148 is merged in. I've approved it, but @jimmidyson needs to finish up some OpenShift template stuff before I can merge it and, subsequently, this one in.

jimmidyson:
add to whitelist

PR #153, opened 2017-02-13
Add pipes for filtering/sorting, wire up toolbar

gashcrumb:
This adds additional filters to the common module for filtering and sorting arrays of objects.  I've gone and wired up the toolbar for the connection list so that you can filter connections again and change the sort order.

Note I've commented out the `type` filter/sort as that property isn't currently defined on a connection object, we'll need to revisit exactly which fields the user should be able to filter and sort on.


src/app/connections/list-toolbar/list-toolbar.component.html
```diff
@@ -11,11 +11,18 @@
...
-              <li><a href="#">Type</a></li>
+              <li><a (click)="setPropertyName('name')">Name</a></li>
+              <!--
+                TODO 'type' is not a valid property currently
+              <li><a (click)="setPropertyName('type')">Type</a></li>
+              -->
             </ul>
           </div>
-          <input type="text" class="form-control" placeholder="Filter By Name..." autocomplete="off" id="filterInput">
+          <input #filterInput
```
kahboom: So for now we are not moving this to a `shared` component, I presume? We obviously have priorities but I would really like to have this available as a reusable component from a common area.

src/app/integrations/create-page/select-connection/select-connection.component.html
```diff
@@ -1,3 +1,3 @@
 <p>Choose a connection to start</p>
-<ipaas-connections-list-toolbar></ipaas-connections-list-toolbar>
-<ipaas-connections-list [connections]="connections | async" [loading]="loading | async" (onSelected)="onSelected($event)"></ipaas-connections-list>
\ No newline at end of file
+<ipaas-connections-list-toolbar  [(filter)]="filter" [(sort)]="sort"></ipaas-connections-list-toolbar>
+<ipaas-connections-list [connections]="connections | async | objectPropertyFilter:filter | objectPropertySort:sort" [loading]="loading | async" (onSelected)="onSelected($event)"></ipaas-connections-list>
```
kahboom: `yarn lint` let you get away with ending a file without a newline? I thought we had `eofline` enabled ([ref](https://palantir.github.io/tslint/rules/eofline/)) [here](https://github.com/redhat-ipaas/ipaas-ui/blob/master/tslint.json#L12). @jimmidyson to the rescue

src/app/connections/list-toolbar/list-toolbar.component.html
gashcrumb: Well, first things first, we've filters with this PR :-)

src/app/integrations/create-page/select-connection/select-connection.component.html
gashcrumb: Updated, think the linter doesn't look at `.html` files.

PR #154, opened 2017-02-14
Add a service for the current integration, use events to communicate between views

gashcrumb:
Basically at least lets you click though some of the flow, i.e. picking a connection, viewing what ought to be the configuration form, then picking the end connection etc.

More work to do, going to bring in [this library](https://github.com/udos86/ng2-dynamic-forms) to render the configuration form for the connection which will replace the json in the second page.

More validation between steps is needed too, side-bar should eventually start being more dynamic.

Also still have to implement saving the integration.  Anyhoo, screencast of how it's looking:

ezgif com-20910625b5

### DIFF
--- a/src/app/integrations/create-page/create-page.component.html
+++ b/src/app/integrations/create-page/create-page.component.html
@@ -1,8 +1,7 @@
 <div class="integrations create">
   <div class="wizard-pf-row">
     <div class="wizard-pf-sidebar">
-      <ipaas-integrations-flow-view 
-        [integration]="integration"></ipaas-integrations-flow-view>
+      <ipaas-integrations-flow-view></ipaas-integrations-flow-view>
     </div>
     <div class="wizard-pf-main">
       <ol class="breadcrumb">
@@ -13,7 +12,10 @@
       <router-outlet></router-outlet>
       <!-- Wizard Footer -->
       <div class='wizard-pf-footer'>
-        <button type='button' class='btn btn-default btn-cancel wizard-pf-cancel wizard-pf-dismiss'>Cancel</button>
+        <button type='button' class='btn btn-default btn-cancel wizard-pf-cancel wizard-pf-dismiss' (click)="cancel()">Cancel</button>
+        <button type="button" class="btn btn-primary" [disabled]="!canContinue" (click)="continue()" *ngIf="!(getCurrentChild() === 'connection-configure' && position === 1)">Next</button>
+        <!-- TODO hiding showing hack -->        
+        <button type="button" class="btn btn-primary" *ngIf="getCurrentChild() === 'connection-configure' && position === 1" (click)="finish()">Finish</button>
       </div>
     </div>
   </div>
--- a/src/app/integrations/create-page/create-page.component.ts
+++ b/src/app/integrations/create-page/create-page.component.ts
@@ -1,11 +1,11 @@
 import { Component, OnInit, OnDestroy } from '@angular/core';
-import { ActivatedRoute, Params, Router } from '@angular/router';
+import { ActivatedRoute, Params, Router, UrlSegment } from '@angular/router';
 import { Observable } from 'rxjs/Observable';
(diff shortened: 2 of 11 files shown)


jimmidyson:
Looking good!

Test failure:

```
IntegrationsCreateComponent should create - Chrome_54_0_2840_(Linux_0_0_0).IntegrationsCreateComponent
Error: No provider for CurrentFlow!
```

gashcrumb:
Hey look, CI works :-)

Lemme fix that

src/app/integrations/create-page/create-page.component.html
```diff
@@ -13,8 +12,11 @@
       <router-outlet></router-outlet>
       <!-- Wizard Footer -->
       <div class='wizard-pf-footer'>
-        <button type='button' class='btn btn-default btn-cancel wizard-pf-cancel wizard-pf-dismiss'>Cancel</button>
+        <button type='button' class='btn btn-default btn-cancel wizard-pf-cancel wizard-pf-dismiss' (click)="cancel()">Cancel</button>
+        <button type="button" class="btn btn-primary" [disabled]="!canContinue" (click)="continue()" *ngIf="!(getCurrentChild() === 'connection-configure' && position === 1)">Next</button>
```
jimmidyson: Not too bothered for now, but would be better to make this a function in the component and call it here.

src/app/integrations/create-page/create-page.component.html
```diff
@@ -13,8 +12,11 @@
...
+        <!-- TODO hiding showing hack -->        
+        <button type="button" class="btn btn-primary" *ngIf="getCurrentChild() === 'connection-configure' && position === 1" (click)="finish()">Finish</button>
```
jimmidyson: Same here - try to limit logic in templates,

gashcrumb: absolutely, this template is still in the POC phase, I'd  expect to be reworking this, hence the TODO in there...

src/app/integrations/create-page/create-page.component.ts
```diff
@@ -20,27 +20,96 @@ export class IntegrationsCreatePage implements OnInit, OnDestroy {
...
+
+  finish() {
+    this.router.navigate(['/integrations']);
+  }
+
+  continue() {
+    const child = this.getCurrentChild();
+    switch (child) {
+      case 'connection-select':
+        this.router.navigate(['connection-configure', this.position], { relativeTo: this.route });
```
jimmidyson: I'm not sure these need to be separate URLs as they will never be bookmarked or navigate to directly, but again something we can address later unless it's a simple fix now.

src/app/integrations/create-page/create-page.component.ts
```diff
@@ -20,27 +20,96 @@ export class IntegrationsCreatePage implements OnInit, OnDestroy {
...
+        break;
+      case 'connection-configure':
+        // TODO hard-coding this to just go to the next connection
+        this.router.navigate(['connection-select', this.position + 1], { relativeTo: this.route });
```
jimmidyson: Same comment re: URLs. I'd prefer to use location hash to direct/link to place in a flow, just like hash pointing to location in HTML document, but again something we can defer.

jimmidyson:
Cool going to [merge] this so you can keep moving forward.

src/app/integrations/create-page/create-page.component.ts
gashcrumb: I wasn't certain if that would be the case, so I wanted to cater for it, I can certainly envision going directly to the connection edit page in an integration for example, etc.

src/app/integrations/create-page/create-page.component.ts
gashcrumb: Yeah, I can look at that vs using encoding the data as part of the route, currently it's used to pick which connection in the integration those controllers should look at.

PR #155, opened 2017-02-14
Add form support and display form on configure connection page

gashcrumb:
Pulled in `ng2-dynamic-forms` for rendering forms and added a service to convert our form config JSON to the objects it uses, end result:

capture


jimmidyson:
Don't know what's up with circle... There's still styling that needs doing but to make sure this doesn't block you, let's merge and look at it later.

PR #156, opened 2017-02-15
UI stuff

kahboom:
Changes:
- Hide things that aren't supposed to be in this sprint.
- Improve things that are supposed to be in this sprint.
- Add collapse to sidebar menu.

No time for unit or E2E tests, but will try to get some in before the demo while I finish working on some styling and the Detail view.

Note: Pulling in the latest changes completely broke my ability to run the app, so it's a good thing I did it after I was done with my work.

screenshot 2017-02-14 21 15 06

A good reminder that I should always rebase against a test branch. I'll likely be bothering @gashcrumb & @jimmidyson about the breaking changes tomorrow. :)

### DIFF
--- a/src/app/integrations/create-page/flow-view/flow-view.component.ts
+++ b/src/app/integrations/create-page/flow-view/flow-view.component.ts
@@ -22,14 +22,13 @@ export class FlowViewComponent implements OnInit, OnDestroy {
   urls: UrlSegment[];
   currentPosition: number;
   currentState: string;
+  isCollapsed: boolean = false;
 
   constructor(
     private currentFlow: CurrentFlow,
     private route: ActivatedRoute,
     private router: Router,
-  ) {
-
-  }
+  ) {}
 
   getClass(state, position) {
     if (state === this.currentState && position === this.currentPosition) {
@@ -65,13 +64,11 @@ export class FlowViewComponent implements OnInit, OnDestroy {
     this.flowSubscription.unsubscribe();
   }
 
-  public isCollapsed: boolean = false;
-
-  public collapsed(event: any): void {
+  collapsed(event: any): void {
     log.debugc(() => 'Event' + event);
   }
 
-  public expanded(event: any): void {
+  expanded(event: any): void {
     log.debugc(() => 'Event' + event);
   }
 
--- a/src/app/integrations/create-page/flow-view/flow-view.component.ts
+++ b/src/app/integrations/create-page/flow-view/flow-view.component.ts
@@ -65,13 +65,13 @@ export class FlowViewComponent implements OnInit, OnDestroy {
     this.flowSubscription.unsubscribe();
   }
 
-  public isCollapsed:boolean = false;
+  public isCollapsed: boolean = false;
 
-  public collapsed(event:any):void {
+  public collapsed(event: any): void {
     log.debugc(() => 'Event' + event);
   }
 
-  public expanded(event:any):void {
+  public expanded(event: any): void {
     log.debugc(() => 'Event' + event);
   }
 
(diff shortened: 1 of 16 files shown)


kahboom:
Should probably squash those commits as well.

kahboom:
Btw, this isn't everything- some more improvements to be made to the CSS but I'm trying to prioritize for the demo. Also, didn't test linting or tests. Sue me.

gashcrumb:
@kahboom test failure on CI:

```
IntegrationsCreateComponent should create - Chrome_54_0_2840_(Linux_0_0_0).IntegrationsCreateComponent

Can't bind to 'collapse' since it isn't a known property of 'ul'. ("  <ul (collapsed)="collapsed($event)"
```

probably just need to import the `ng2-bootstrap` collapse module into `create-page.component.spec.ts`

kahboom:
@gashcrumb Thanks! Were you still getting that error I got yesterday when testing this branch? I fixed it about 10 mins after I messaged you yesterday so it should be working.

gashcrumb:
@kahboom did you push your fix?  It doesn't look fixed in `create-page.component.spec.ts` looking here...

kahboom:
@gashcrumb - I had to import it from the integrations module, because doing it from the root `app.module.ts` file wasn't working. It's not done at the component level (unless you want that changed?). See here: https://github.com/redhat-ipaas/ipaas-ui/pull/156/files#diff-0da558fb66c7be27c703bb77dcd59c20R18

That took away the error, but then pulling in the latest changes results in a CORS error as well as an error related to a relative path leading to a 404 (cc @jimmidyson ) as shown below, using the remote API (running locally doesn't seem to work for me either, perhaps for a different reason), which breaks the UI completely:

screenshot 2017-02-15 08 41 56

gashcrumb:
@kahboom gotcha.  Yeah, CORS issue is due to the backend, don't worry about that.

Even though we import that module into the root of the app you still actually have to import it into the Integrations module as well.  Note that you only use the `forRoot()` function when importing it into the app module, when you're importing it into a child module you don't need that.

And for the test, you have to actually add all the same code as in the integrations module to the `.spec.ts` file, as the test basically creates a stripped down module that only contains that component and it's dependent services/modules.

src/app/integrations/integrations.module.ts
```diff
@@ -40,6 +41,7 @@ const routes: Routes = [
     ConnectionsModule,
     TabsModule,
     IPaaSCommonModule,
+    CollapseModule.forRoot(),
```
gashcrumb: Don't use `.forRoot()` here, that should only be used in `app.module.ts`

kahboom: Ah I kind of figured but didn't take the time to google it. :/

src/app/integrations/create-page/flow-view/flow-view.component.ts
```diff
@@ -65,4 +65,14 @@ export class FlowViewComponent implements OnInit, OnDestroy {
     this.flowSubscription.unsubscribe();
   }
 
+  public isCollapsed:boolean = false;
+
+  public collapsed(event:any):void {
+    console.log(event);
```
gashcrumb: this passed linting?

kahboom: hehe.. that's what happens when you copy and paste sample code from an API. Fixing now.

kahboom:
@gashcrumb - Apologies for the delay, fixed the linting errors, unit tests, cleaned up logs, removed `.forRoot()` for the CollapseModule. Trying to think if I missed anything... hrm.

gashcrumb:
@kahboom no worries!  Took a look, looks great, just waiting for CI to finish up...

PR #160, opened 2017-02-15
Dynamically set page title, add back button, tweak connection selection

gashcrumb:
This PR ensures the page title in the integration editor matches the designs as you go through the wizard.

Also added the back button, though currently you can't go back once you've set the start connection currently.  Just shhh don't tell anybody that.

Also added chevrons and s/Finish/Create/ per the design.


kahboom:
Failed a lint test for trailing whitespace, but I'll fix that in my next PR.

PR #164, opened 2017-02-16
Create an Integration UI Fixes

kahboom:
Changes:
- Aesthetic changes for Create an Integration view to reflect designs more closely. (e.g. fix layout issue, sidebar)
- Fix lint error caused by previous PR merge w/ trailing whitespace.

Again, apologies for not squishing commits- it's crunch time. :)

### Create an Integration

screenshot 2017-02-16 08 52 21

In Progress (next PR, all minor tweaks):
- Center + icon in cube-line design on left-hand side
- Integration List improvements
- Connection Detail improvements

Low priority (may not happen for the demo):
- Automatically collapse the sidebar when in the Create Integration view, otherwise the layout is too wide and you can scroll to the right (for the demo we may have to manually collapse it).
- Look into issue where child components don't inherit CSS from parent components located in different modules.
- I might adjust the bottom bar of the Create Integration page to reflect designs to be gray and within the list (right now it kind of overlaps the sidebar), but since this will be moved to the top toolbar in the new designs it's likely not worth fixing at all, especially since it looks okay as is.
- Hover effect of steps in sidebar? Not sure if this was a thing, can ask UX.

### DIFF
--- a/src/app/integrations/create-page/create-page.component.html
+++ b/src/app/integrations/create-page/create-page.component.html
@@ -8,7 +8,9 @@
         <li><a [routerLink]="['/']"><i class="fa fa-angle-double-left"></i> Dashboard</a></li>
         <li class="active">{{ pageTitle }}</li>
       </ol>
-      <h1>{{ pageTitle }}</h1>
+      <div class="title">
+        <h1>{{ pageTitle }}</h1>
+      </div>
       <router-outlet></router-outlet>
       <!-- Wizard Footer -->
       <div class='wizard-pf-footer'>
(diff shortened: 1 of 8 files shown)


kahboom:
Also need to fix width of the cards, but that's also related to the inheritance problem. Might look into that sooner than later so that things look nice for the demo, time-permitting.

kahboom:
For reference, this is the design: https://redhat.invisionapp.com/share/VQ8UZ7PCE#/198707867_Create_Integration_-_From_Scratch

jimmidyson:
Thanks! [merge]

PR #168, opened 2017-02-16
Fix store service error exception, display connection fields on conne…

gashcrumb:
…ction view page, fix up page flow from integration editor

This fixes the connection view page to show the available fields and their settings, also the store services were erroring out when getting HTTP error returns, this sorts that so views can handle errors.

Also this ensures the create integration flow transitions back to the integration list when you click create, note we don't have toaster notifications in place yet, so we don't yet have a decent way to notify the user if the integration was created or if there was an error.


kahboom:
I'll try to get the toast notifications in ASAP. Thanks!

PR #174, opened 2017-02-17
Generate models from swagger

gashcrumb:
This fixes https://github.com/redhat-ipaas/ipaas-rest/issues/80

It updates the code to use the generated model and adds a new script in `package.json` to run the generator:

`yarn run generate`

Which we can run whenever the model is updated for the rest API.  Ideally at some point I'd love to figure out how we could automate the code generation via CI, but for now manually works.  Until it gets painful :-)


jimmidyson:
Thanks! [merge]

Issue #114, opened 2017-01-14
support websockets in the RestService/Store abstraction so we can update entities in real time


jstrachan:
we're used in the fabric8 console to having everything using websockets so that we can see quickly when things change without having to poll.

e.g. of 2 folks are working in the same namespace; a Java dev may install some Connectors which then should appear immediately in the Citizen's UI etc.

It'd be nice to at least conceptually support this feature.

I wonder if the use of Observable<Array<T>> is the right abstraction in our underlying models? i.e. if one Foo updates, does that mean we'd have a new Array<T> event and so force a redraw an entire table?

jimmidyson:
If we get an update for a single entity via websockets, then we would replace entity in store array & emit event with full array. We should probably use track by to ensure only changed elements are redrawn rather than whole list.

chirino:
This is done.. events can be obtained via SSE or Websockets.

PR #186, opened 2017-02-22
Use the new Server Side Event support in ipaas-rest to automatically reload updated entity lists.

chirino:
Works in conjunction with the following PR:
https://github.com/redhat-ipaas/ipaas-rest/pull/115

### DIFF
--- a/src/app/store/entity/entity.store.ts
+++ b/src/app/store/entity/entity.store.ts
@@ -1,4 +1,6 @@
 import { Observable } from 'rxjs/Observable';
+import {Observer} from 'rxjs/Observer';
+
 import { BehaviorSubject } from 'rxjs/BehaviorSubject';
 import { Subject } from 'rxjs/Subject';
 import { plural } from 'pluralize';
@@ -7,6 +9,7 @@ import { RESTService } from './rest.service';
 import { BaseEntity } from '../../model';
 
 import { log, getCategory } from '../../logging';
+import {EventsService, ChangeEvent} from './events.service';
 
 const category = getCategory('AbstractStore');
 
@@ -19,18 +22,39 @@ export abstract class AbstractStore<T extends BaseEntity, L extends Array<T>,
 
   private _loading: BehaviorSubject<boolean> = new BehaviorSubject(false);
 
-  constructor(private service: R, initialList: L, initialCurrent: T) {
-    this._list = new BehaviorSubject(initialList);
-    this._current = new BehaviorSubject(initialCurrent);
+  private changeEvents: Observable<ChangeEvent>;
+
+  constructor(private service: R, private eventService: EventsService, initialList: L, initialCurrent: T) {
+    this._list = new BehaviorSubject<L>(initialList);
+    this._current = new BehaviorSubject<T>(initialCurrent);
+
+    this.changeEvents = this.eventService.changeEvents.filter((x) => {
+      return x.kind === this.service.kind;
+    });
+
   }
 
   protected abstract get kind(): string;
 
-  get list() { return this._list.asObservable(); }
+  get list() {
+    // Give back the _list,
+    // but also update it if we get notified the a change occurred.
+    return Observable.merge(
+      this._list,
+      this.changeEvents.flatMap((event) => {
+        // We could probably get fancy one day an only fetch the entry that matches event.id
+        return this.service.list();
+      }),
+    ).share();
+  }
 
-  get resource() { return this._current.asObservable(); }
+  get resource() {
+    return this._current.asObservable();
+  }
 
-  get loading() { return this._loading.asObservable(); }
+  get loading() {
+    return this._loading.asObservable();
+  }
 
   loadAll() {
     this._loading.next(true);
@@ -105,11 +129,11 @@ export abstract class AbstractStore<T extends BaseEntity, L extends Array<T>,
   }
 
   /*
-  deleteEntity(id?: string) {
-    if(id) {
-      this.service.delete(id);
-    }
-  }
-  */
+   deleteEntity(id?: string) {
+   if(id) {
+   this.service.delete(id);
+   }
+   }
+   */
 
 }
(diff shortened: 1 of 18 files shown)


jimmidyson:
@dsimansk @jludvice Stuff like this EventSource usage highlights the need for cross-browser test suites... Please have a think about how we can do that.

chirino:
retest this please

jimmidyson:
Does EventSource support authorization headers? Our REST API currently only authenticates via validation of JWT contained in authorization header...

chirino:
@jimmidyson Looks we have to pass it via cookies.. is that a deal breaker for SSE?

src/app/store/entity/entity.store.ts
```diff
@@ -19,18 +22,39 @@ export abstract class AbstractStore<T extends BaseEntity, L extends Array<T>,
...
+      this._list,
+      this.changeEvents.flatMap((event) => {
+        // We could probably get fancy one day an only fetch the entry that matches event.id
+        return this.service.list();
+      }),
+    ).share();
+  }
 
-  get resource() { return this._current.asObservable(); }
+  get resource() {
```
kahboom: Not sure if this was you or `ts-lint.json`, but either way, ty.

kahboom:
If something doesn't make sense in my comment, just know I'm running on 3 hours of sleep, which may or may not be relevant. ;D

kahboom:
Even better, a Node.js implementation that can be used to override headers from the "JS side", polyfill included: [eventsource](https://github.com/EventSource/eventsource)

chirino:
@kahboom, @jimmidyson I think SSE security should be sorted now without needing to do the cookie thing.  Basically I first do a POST to request a event stream ID, at that point the user is associated with the stream, and then then we setup the SSE connection against that registered stream id (a random UUID) which does not get re-used.

I kind like the SSE first then fallback to WS scenario, since if we ever move to HTTP2, then we are using the preferred strategy to begin with. 

also take a peek at:
http://caniuse.com/#feat=eventsource
http://caniuse.com/#feat=websockets

Notice we would be using SSE on most browsers except IE and Edge, and both have good WebSocket support.

kahboom:
Yeah Websockets have the advantage of wider browser support. I'm not opposed to falling back on WS, or at least don't feel strongly about it. @jimmidyson I'd like to hear your thoughts on this. @chirino once tests are resolved I'll approve it, then merge based on what Jimmi prefers.

jimmidyson:
add to whitelist

src/app/store/entity/events.service.ts
```diff
@@ -0,0 +1,90 @@
...
+  messageEvents: Subject<String> = new Subject<String>();
+  changeEvents: Subject<ChangeEvent> = new Subject<ChangeEvent>();
+
+  constructor(config: ConfigService, restangular: Restangular) {
+    // If we have trouble setting up the EventSource
+    if (!this.connectEventSource( restangular, config.getSettings().apiEndpoint )) {
+      // , fallback to using a WebSocket
+      let wsurl = config.getSettings().apiEndpoint;
+      // I know hacky!  need to figure out how to put the wsevents resource under the apiEndpoint
+      wsurl = wsurl.replace( /^https?/, 'ws' ).replace( /\/api\/v\d+$/, '/wsevents' );
```
jimmidyson: this would always fall back to insecure ws rather than wss if https is used

kahboom:
Providing Websocket support as a fallback means we would need to maintain it as the application scales, otherwise it'd be for naught. Just something to consider, but again, not opposed to it at all.

src/app/store/entity/events.service.ts
```diff
@@ -0,0 +1,90 @@
+import {Injectable} from '@angular/core';
+import {Subject} from 'rxjs/Subject';
+import {UUID} from 'angular2-uuid';
```
jimmidyson: This doesn't seem to be used?

jimmidyson:
Can't remember if the server side even supports websockets right now - @chirino?

src/app/store/entity/events.service.ts
kahboom: Right, we'd absolutely need WSS considering we'd be opening connections quite frequently if it's to update models on any change.

kahboom:
@jimmidyson - Now it does I believe, see https://github.com/redhat-ipaas/ipaas-rest/pull/115

kahboom:
Also of relevance #114 .

src/app/store/entity/events.service.ts
chirino: yep.. don't need the uuid anymore.  hangover from previous iteration.

src/app/store/entity/events.service.ts
chirino: will fix in a bit.

chirino:
Squashed and rebased off master.

chirino:
@kahboom, @jimmidyson: yep ipaas-reset supports both SSE and WebSockets.

src/app/store/entity/events.service.ts
```diff
@@ -0,0 +1,89 @@
...
+  messageEvents: Subject<String> = new Subject<String>();
+  changeEvents: Subject<ChangeEvent> = new Subject<ChangeEvent>();
+
+  constructor(config: ConfigService, restangular: Restangular) {
+    // If we have trouble setting up the EventSource
+    if (!this.connectEventSource( restangular, config.getSettings().apiEndpoint )) {
+      // , fallback to using a WebSocket
+      let wsurl = config.getSettings().apiEndpoint;
+      // I know hacky!  need to figure out how to put the wsevents resource under the apiEndpoint
+      wsurl = wsurl.replace( /^http/, 'ws' ).replace( /\/api\/v\d+$/, '/wsevents' );
```
jimmidyson: hmm this is a bit pants and means we need to set up another subpath route in openshift. what's the problem putting it under the API endpoint?

chirino: Yep that's what I initially tried but the websocket handler would not trigger if not at a top level path. Not sure why yet.  Perhaps if I use a different WS handling strategy it might work better (I'm currently using the SB abstractions).

jimmidyson:
@kahboom Are you happy with this? LGTM

PR #193, opened 2017-02-28
Two phase OAuth flow to support OpenShift and GitHub

jimmidyson:
I'd like to merge this ASAP but this will break any local use due to the single
redirect policy for GitHub OAuth apps. Would appreciate help coming to right
solution so we don't break everyone.

3 potential solutions:

1. Each dev have their own oauth app pointing at whatever local redirect URL they
want. This is my preferred option to be honest as it's most flexible but is also
hardest to set up for new contributors.

2. Each dev use the same URL locally and a single oauth application. This would
work but coming up with a URL that works for everyone might prove tricky. Also
have to share oauth app secrets somehow. Not good for new contributors perhaps.

3. Set up a @chirino's redirector that can be used to redirect to local app. Not sure I like
this as it goes against normal oauth flow.

We can always do something for now and come up with a better solution later.

Thoughts @gashcrumb @kahboom @chirino ?

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,7 +27,7 @@
     "@angular/router": "3.4.7",
     "@ng2-dynamic-forms/core": "1.3.13",
     "@ng2-dynamic-forms/ui-bootstrap": "1.3.13",
-    "angular-oauth2-oidc-hybrid": "1.0.22",
+    "angular-oauth2-oidc-hybrid": "1.0.26",
     "angular2-toaster": "2.0.0",
     "bootstrap": "3.3.7",
     "core-js": "2.4.1",
--- a/src/app/app.module.ts
+++ b/src/app/app.module.ts
@@ -25,8 +25,6 @@ import { log } from './logging';
 export function appInitializer(configService: ConfigService, oauthService: OAuthService, userService: UserService) {
   return () => {
     return configService.load().then(() => {
-      // URL of the SPA to redirect the user to after login
-      oauthService.redirectUri = window.location.origin + '/dashboard';
       oauthService.clientId = configService.getSettings('oauth', 'clientId');
       oauthService.scope = (configService.getSettings('oauth', 'scopes') as string[]).join(' ');
       oauthService.oidc = configService.getSettings('oauth', 'oidc');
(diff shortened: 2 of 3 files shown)


gashcrumb:
@jimmidyson we don't really have an easy way to set the local dev instance's host/port anyways, so I don't see that option 2 is that bad.  When we do, then I can't imagine supporting option 1 at that point would be difficult, would it?

kahboom:
The only time it'd be tricky for new contributors is when things are not documented properly, so I think if you prefer option 1, we should stick with it and document it well (everywhere; ie on the website, the README, share it with the technical writers, on the blog once we have one, etc.). You'd be surprised by how smart users are when provided with detailed documentation that explains what they are doing and why, perhaps presenting them with an alternative and its benefits, etc.

I personally don't like the idea of sharing oauth app secrets (option 2); wouldn't that be unsustainable and insecure, or am I misunderstanding something here?

gashcrumb:
>I personally don't like the idea of sharing oauth app secrets (option 2); wouldn't that be unsustainable and insecure, or am I misunderstanding something here?

That'd be our organization's client ID key I think, which yeah, that might get ugly.  I mean, we could each set it locally via an environment variable or put it in our local config.json copy.  I guess, how much external contributions should we expect at this stage?

jimmidyson:
I expect exactly zero external contributors :sob:.

It would require local updates to config.json, sharing secrets via email for now probably. We should also be able to use our staging keycloak as you are now probably and set up redirect to lcoalhost:4200 for local hacking. That should work nicely actually.

jimmidyson:
I'll update docs so people can use option 2 for local hacking (http://localhost:4200) and 1 for running in a cluster where they will need to be able to set up a "real" github oauth app. Let me do that before merging.

Any comments on code btw?

kahboom:
>I expect exactly zero external contributors 😭.

Well, with that attitude. ;D

We need to focus on marketing for upstream; I'd need technical help as I'm technically disabled (not "technical" in the sense of "truly", rather "knowledge-wise", also I'm not "disabled", more like "impaired" .. or "hindered".. yes let's go with "hindered"). Things like SEO and community-building take forever to start having an impact.

> I'll update docs so people can use option 2 for local hacking

👍 

> Any comments on code btw?

That's how much trust I have in you Jimmi; I'm not even going to look.

PR #195, opened 2017-03-01
Fix NPE

gashcrumb:
This error's been bugging me for a couple weeks, let's hide it...


jimmidyson:
add to whitelist

jimmidyson:
Thanks @gashcrumb! Very annoying indeed. It's actually a bug in the REST API sending back empty response... Should fix that properly too!

[merge] fo rnow

PR #198, opened 2017-03-02
Create connection flow step 2

gashcrumb:
This gets page 2 of the create connection flow going.

No data from the form is saved yet, probably not going to happen until we've ironed out the persistence.  Also all pages in the wizard require styling work, hoping to get that going tomorrow.

Obligatory screenshot:

screencapture-localhost-4200-connections-create-configure-fields-1488489416370


jimmidyson:
Let's get this [merge]d!

PR #199, opened 2017-03-03
Read/save values into the config string

gashcrumb:
For now (this will all need adjusting) we'll save/read values from 'configuredProperties', so now connections actually save their values, for example:

screencapture-localhost-4200-connections-6-1488555766545


jimmidyson:
Thanks! [merge]

PR #208, opened 2017-03-07
Refactor the page title/breadcrumbs to individual wizard pages, fix #207

### LINKED ISSUE #207: Refactor page title/wizard body portion of integration editor wizard
We should move this block -> https://github.com/redhat-ipaas/ipaas-ui/blob/master/src/app/integrations/create-page/create-page.component.html#L7-L14

into the individual wizard page so we don't have to do stuff like this -> https://github.com/redhat-ipaas/ipaas-ui/blob/master/src/app/integrations/create-page/create-page.component.ts#L140-L153

@kahboom will take a look at doing this tomorrow...

gashcrumb:


jimmidyson:
LGTM thanks! [merge]

PR #210, opened 2017-03-07
Regen models, add action service/store

gashcrumb:
Adds in the new action type and associated restangular service/store, pointing at the top-level 'actions' API

### DIFF
--- a/src/app/model.ts
+++ b/src/app/model.ts
@@ -25,16 +25,25 @@ export interface WithId extends BaseEntity {
 };
 export type WithIds = Array < WithId > ;
 
+export interface Action extends BaseEntity {
+    description: string;
+    connectorId: string;
+    properties: string;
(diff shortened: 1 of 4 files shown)


gashcrumb:
Hold off merging this, want to maybe look at fetching the actions via `/connector/{id}/actions`

gashcrumb:
Hmm, actually never mind, merge at will :-)

src/app/model.ts
```diff
@@ -139,24 +132,34 @@ export type Tags = Array < Tag > ;
...
     roleId: string;
+    organizationId: string;
     name: string;
     id: string;
 };
 export type Users = Array < User > ;
 
+export interface ListResultAction extends BaseEntity {
+    items: Array < Action >
+    ;
```
kahboom: What? :)

PR #216, opened 2017-03-07
Use nav management functions to show/hide the main nav where appropriate

gashcrumb:
Since we got the PR in to show/hide the main nav programatically we can now use this in the create connection and create integration controllers to hide the main nav when the user goes into those pages.


kahboom:
@gashcrumb - Tests are currently failing.

PR #223, opened 2017-03-09
Fix action select and configure pages, fix up how we're saving stuff …

gashcrumb:
…into the integration


kahboom:
> As an administrator, you may still merge this pull request.

POWER! Unlimited POWERRRRR! https://www.youtube.com/watch?v=kNS4t5UCBfI

PR #226, opened 2017-03-09
Move buttons to the top-right, refactor page flow

gashcrumb:


src/app/integrations/edit-page/action-configure/action-configure.component.html
```diff
@@ -1,15 +1,32 @@
-<ol class="breadcrumb">
-  <li><a [routerLink]="['/']"><i class="fa fa-angle-double-left"></i> Dashboard</a></li>
-  <li class="active">Configure {{ action.name }}</li>
-</ol>
-<div class="title">
-  <h1>Configure {{ action.name }}</h1>
+<!-- Toolbar -->
+<div class="toolbar">
```
kahboom: Thank god we are doing this.. Honestly, keeping code DRY is great practice, but for right now we need to keep things in scope, including for styling. Plus we have a demo tomorrow rofl.

PR #229, opened 2017-03-09
Add links to insert steps/connections

gashcrumb:
This adds the links that let you add a step or connection to a flow.  They only show up on this page.  I think BTW that the buttons in the main view make no sense as there's no context to figure out where a step/connection should be added in the flow, unlike the inlined links.  Something to discuss with UX.

screencapture-localhost-4200-integrations-edit-2-save-or-add-step-1489096269977

### DIFF
--- a/src/app/integrations/edit-page/edit-page.component.ts
+++ b/src/app/integrations/edit-page/edit-page.component.ts
@@ -1,4 +1,4 @@
-import { Component, OnInit, OnDestroy, AfterViewInit } from '@angular/core';
+import { Component, OnInit, OnDestroy, ChangeDetectorRef } from '@angular/core';
 import { ActivatedRoute, Params, Router, UrlSegment } from '@angular/router';
 import { Observable } from 'rxjs/Observable';
 import { Subscription } from 'rxjs/Subscription';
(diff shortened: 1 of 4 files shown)


kahboom:
cc @sjcox-rh @dongniwang just as a head start for next sprint ;D

src/app/integrations/edit-page/edit-page.component.ts
```diff
@@ -72,6 +73,7 @@ export class IntegrationsEditPage implements OnInit, OnDestroy {
     switch (event.kind) {
       case 'integration-updated':
         this.router.navigate(['save-or-add-step'], { relativeTo: this.route });
+        this.detector.detectChanges();
```
kahboom: hehe.. gotta love `detectChanges()`.

src/app/integrations/edit-page/flow-view/flow-view.component.html
```diff
@@ -21,17 +20,27 @@
           [currentPosition]="currentPosition"
           [currentState]="currentState"></ipaas-integrations-flow-view-step >
       </div>
-      <div class="row steps" *ngFor="let step of getMiddleSteps(); let position = index;">
-        <div *ngIf="currentPosition === 'new'">
-          New</div>
+      <div class="row steps" *ngIf="currentState === 'save-or-add-step'">
```
kahboom: I love this; much, much easier to understand and easier to see right away when looking at the markup. Usually not a fan of having logic in the markup, but this is really minimal.

kahboom:
Wait this makes the Create Integration page default to the middle steps ('save-or-add-step') and breaks the whole creation flow. I would revert the merge but since I don't know if you have a local copy of that branch I'll just work around it. I love this but not the breakage lol. Please fix ASAP @gashcrumb , thanks! :)

jimmidyson:
Please!

Issue #214, opened 2017-03-07
Integration List Toolbar Sort & Filter


kahboom:
This is currently working for Connections, but not for Integrations.

PR #230, opened 2017-03-09
re #214 Fix for Integration List Search & Sort

kahboom:
### Overview
Fixes the toolbar for the Integration List, so that you can search and sort items properly.

[Screen capture](https://www.useloom.com/share/69d0da2d555a4b73809d096ced16c794)

Why did I take so long to do this? :)


jimmidyson:
LGTM thanks!

PR #238, opened 2017-03-10
Split out steps into their own store service, switch add step/connection links to use a popover

gashcrumb:


kahboom:
Thanks! That was exactly what I was doing, making them into a popover but keeping them opened permanently for this step: https://redhat.invisionapp.com/share/VQ8UZ7PCE#/222155519_Save_Or_Add_Step

Issue #266, opened 2017-03-16
Create Integration: Add Filter Step


kahboom:
## Overview
Add the ability to filter data as a step when creating an integration in the UI. Entails the following:
- Add HTML & SCSS to reflect designs & incorporate logic.
- Add method(s) called by HTML for adding the filter step and making the call to the REST API in its respective component.

### Relevant Designs
N/A

![]()

## Reference
- [InVision: Design]()
- [JIRA Issue: UI](https://issues.jboss.org/browse/IPAAS-270)
- [JIRA User Story](https://issues.jboss.org/browse/IPAAS-245)

gashcrumb:
Step is in place.

Issue #248, opened 2017-03-13
Create Integration - Connection ring color change


sjcox-rh:
Once a user is done with setting up connection, the ring around the connection icon should be blue, not gray. Should remain blue for rest of flow.

<img width="333" alt="screen shot 2017-03-13 at 12 17 36 pm" src="https://cloud.githubusercontent.com/assets/16257540/23871479/d4d0fdac-07e7-11e7-8afe-e4655a56e966.png">

gashcrumb:
Apparently this isn't showing up correctly still

PR #270, opened 2017-03-16
Get the filter stepkind in place for IPAAS-270

gashcrumb:
This adds the filter step to the list, though we've gotta find out what properties are needed for it.


kahboom:
re #248 #266

PR #285, opened 2017-03-17
Don’t add value to the action properties.

chirino:


gashcrumb:
Can you fix the lint error:

> src/app/integrations/edit-page/action-configure/action-configure.component.ts[58, 1]: trailing whitespace

PR #292, opened 2017-03-22
Hide 'choose connection/action/step' once the user has gone through those pages

gashcrumb:
This hides the 'Choose a Connection' and 'Choose an action' links on the left after the user has gone through those pages and defaults the parent item (connection or step name) to link to the 'Configure' page containing the form instead.

@kcbabo @chirino FYI


gashcrumb:
So as an example:
screencapture-localhost-4200-integrations-edit-2-step-configure-2-1490194085923

kcbabo:
Looks good!

Issue #314, opened 2017-03-24
Create Integration: Add to Integration Buttons Nonfunctional


kahboom:
"Rest stop" displayed to users for the first time upon completion of configuring both connections displays buttons in the main content area that are nonfunctional (ie clicking them does not result in an action).

Design:

![save or add step](https://cloud.githubusercontent.com/assets/3844502/24300312/6d7db002-1082-11e7-92ed-a0afccff02cb.jpg)

kahboom:
@gashcrumb - This works w/ the exception of the redirect and forcing the tooltip to stay open (#313). Would you mind taking a look at the redirect issue?

Issue #305, opened 2017-03-22
Fix Disabled State for Steps


kahboom:
Currently looks like this: #notgood

<img width="1440" alt="screenshot 2017-03-22 16 02 11" src="https://cloud.githubusercontent.com/assets/3844502/24218316/2e76d452-0f19-11e7-9b2a-2c8c8c232ddf.png">

Should look like this:

![add step - first step - 03](https://cloud.githubusercontent.com/assets/3844502/24225759/9d8ebdd6-0f38-11e7-8c76-8d6ce48dda5f.jpg)

kahboom:
Icon state classes not currently working properly.

gashcrumb:
This probably has an impact on this issue -> https://github.com/redhat-ipaas/ipaas-ui/issues/337

kahboom:
@gashcrumb - There shouldn't be an icon in the center at all, so it's just a matter of determining when it should be inactive or incomplete and then updating the added classes so that it's gray instead of blue.

gashcrumb:
Yep, that should be in place, i.e. if `stepKind` and `configuredProperties` on the step are set then it should be blue, otherwise gray, unless the user is looking at that step.

gashcrumb:
oh poo, or at least I thought I had those in place, but now it doesn't look right :-(

akieling:
Is this issue still relevant?

gashcrumb:
Nope...

PR #316, opened 2017-03-24
Create Integration Changes

kahboom:
## Changes
- Add space between start and finish icons, require that line take up the full height of the division as it grows and shrinks, with a minimum height of 200px always being required.
- Refactor SCSS for `flow-view-step.component.scss` to accommodate for active/inactive and complete/incomplete state classes.
- Add method (not functional, but that's for a separate PR) to determine whether section is complete or incomplete.
- Fix bug that was causing icons to all have `inactive` states. re #305
- 'Add Step' and 'Add Connection' buttons on first-time landing page when saving or adding a step are functional, though redirect does not work properly. re #314 

## Screenshots

screenshot 2017-03-24 18 12 01

## To Do

- Determine when a Step is complete and assign appropriate class.
- Force tooltip to remain open on first time user is prompted to add a step or connection (should this be a locally stored var to reference from other places?).
- Fix redirect issue with buttons on Save or Add Step/Connection page.

### DIFF
--- a/src/app/integrations/edit-page/flow-view/flow-view-step.component.html
+++ b/src/app/integrations/edit-page/flow-view/flow-view-step.component.html
@@ -30,7 +30,7 @@
   <!-- Icon Progress Line -->
 
   <div class="col-md-3 progress-line">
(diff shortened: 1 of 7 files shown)


gashcrumb:
Looking good!

Issue #245, opened 2017-03-13
align usage of <button> and <a class="button">


jludvice:
We use multiple ways of specyfiing buttons. For instance 
here it's ` <a [routerLink]=" ['/integrations/create'] " class='btn btn-primary`
https://github.com/redhat-ipaas/ipaas-ui/blob/f8b8fd9bff62a5576c21b35e34ebcc96ff6cf514/src/app/dashboard/emptystate.component.html#L13

and here it's  `<button class="btn btn-default" (click)="cancel()">Cancel</button>`

I wote for using `button` because 
* it makes sense semantically
* there is direct support by protractor to search button by it's title 
http://www.protractortest.org/#/api?view=ProtractorBy.prototype.buttonText 

I wouldn't mind to open PR as an excercise if you @gashcrumb @kahboom  agree with this alignment

gashcrumb:
@jludvice yeah, if you've time to do a PR that's absolutely welcome :-)  Thanks!

jludvice:
Opened PR #247

gashcrumb:
ooh, this was merged, closing :-)

jludvice:
Reopening to align integration edit page

jludvice:
PR #320 merged. Closing issue.

PR #320, opened 2017-03-27
[relates #245] Align button usage

jludvice:


gashcrumb:
@jludvice ah, good catch, thanks!

PR #324, opened 2017-03-27
Fix #314, automatically insert the new step, or popout the popovers so the user sees where to click next

### LINKED ISSUE #314: Create Integration: Add to Integration Buttons Nonfunctional

gashcrumb:


gashcrumb:
apply this one first -> https://github.com/redhat-ipaas/ipaas-ui/pull/323

src/app/integrations/edit-page/save-or-add-step/save-or-add-step.component.ts
```diff
@@ -79,6 +79,20 @@ export class IntegrationsSaveOrAddStepComponent extends FlowPage implements OnIn
...
   }
 
+  addNew(type: string) {
+    this.currentFlow.events.emit({
+      kind: 'integration-add-step',
+      type: type,
+    });
+  }
+
+  showPopouts(type: string) {
```
kahboom: How are we handling _which_ popover to show? Or does it not matter because only one will ever be present? We need to test manually for any edge cases.

Issue #300, opened 2017-03-22
Dashboard: Empty State


kahboom:
## Overview
Need to update the Dashboard page so that it shows the proper empty state when no integrations have been created. Should have minimal logic. Entails the following:
- Add HTML & SCSS to reflect designs & incorporate logic.
- Add respective methods in component to display Connections at the bottom, as shown in the design (Connections replaced Templates for now).

### Designs

![dashboard v7-empty](https://cloud.githubusercontent.com/assets/3844502/24213480/f9b62bd8-0f08-11e7-8de9-fb86d34c1251.png)

## Reference
- [GitHub: Dashboard](https://github.com/redhat-ipaas/ipaas-ui/issues/296)
- [GitHub: API Issue](https://github.com/redhat-ipaas/ipaas-rest/issues/246)
- [InVision: Design](https://redhat.invisionapp.com/share/VQ8UZ7PCE#/210743184_Dashboard_V7-Empty)
- [JIRA Issue: UI](https://issues.jboss.org/browse/IPAAS-281)
- [JIRA User Story](https://issues.jboss.org/browse/IPAAS-279)

TovaCohen:
When the expert user creates a connection, is it always considered to be a connection template?
When you create a template, are you always creating a connection template? In other words, is that the only kind of template?
I am wondering if the citizen user is always selecting from a set of connection templates rather than a set of connections. 
On the Connection Templates, what are the Dev and Update boxes for?

kahboom:
Whoops

Issue #296, opened 2017-03-22
Dashboard


kcbabo:
## Overview
Need to update the dashboard page so that it shows the summary view when we have at least one integration deployed. Entails the following:
- Add HTML & SCSS to reflect designs & incorporate logic.
- Add method(s) called by HTML for counting filtered integrations to display accurate data.
- Add method(s) called by HTML to request filtered integrations and connections from the API.

cc @KurtStam @chirino - We need to be provided with properties that include the last date/time an integration was updated and the messages exchanged (count, it will be displayed to the user as "No. of times Used" - see design below for reference). Honestly, the API should accept query params like `/api/v1/integrations?sort=lastUpdated&asc=true&limit=10` if it doesn't already. We could do the sorting and filtering from the UI but it will get messy quickly. It's also probably best practice to do this from the API. The "count" for active, inactive, and draft integrations required by the designs is not a big deal and can be done from the UI if anything.

## Designs

**Dashboard**

![225246638](https://cloud.githubusercontent.com/assets/3844502/24380465/75f44586-1319-11e7-8218-cf98e34fe0ee.png)

![225246460](https://cloud.githubusercontent.com/assets/3844502/24380462/73716294-1319-11e7-95fc-d39eee1862b5.png)

## Reference
- [GitHub: Dashboard: Empty State](https://github.com/redhat-ipaas/ipaas-ui/issues/300)
- [GitHub: API Issue](https://github.com/redhat-ipaas/ipaas-rest/issues/246)
- [InVision Design: Dashboard 01](https://redhat.invisionapp.com/share/VQ8UZ7PCE#/225246638_IPaaS_Dashboard_3-23_P1-1x)
- [InVision Design: Dashboard 02](https://redhat.invisionapp.com/share/VQ8UZ7PCE#/225246460_IPaaS_Dashboard_3-23_P2-1x)
- [JIRA Issue: UI](https://issues.jboss.org/browse/IPAAS-280)
- [JIRA User Story](https://issues.jboss.org/browse/IPAAS-279)

kahboom:
Note: After discussing with @sjcox-rh and @dongniwang, the design will remain the same with the exception of the templates being replaced with Connections. The checkmark icons on the left-hand side represent that status of the integration, which is not currently implemented. For demoing purposes they will be left there (and incorporated into the Integration List view), but will not have any functional purpose until preview.

gashcrumb:
[...]

PR #325, opened 2017-03-28
re #296 #300 Dashboard: Empty & Non-Empty States

kahboom:
## Changes
- Add component directives for connections and integrations for the dashboard.
- Add conditional method to a directive selector that checks length of items in Integrations array to determine whether or not to display the template.
- Load all integrations and connections in the dashboard (this needs to be modified to load only what we need).
- Add card view for separating widgets in the UI.
- Move list of connections loading to the connection list component for the Dashboard.
- Aesthetic changes to resemble design more closely.

## To Do
- Use actual status for Top Integrations.
- Use actual data & label for Recent Updates.
- Tooltip/dropdown is not working properly for Connection section.
- Sort and modify list if the API does not support this already.
- Properly integrate chart to display Integration Board (and use actual data).


kahboom:
cc @gashcrumb - One test failing. Still working on this (see below), but once it's resolved (any ideas?) feel free to review and merge.

screenshot 2017-03-28 21 13 15

kahboom:
I'll need to create a second PR for the rest as it's too much to put into one PR and this one is already yuuuge.

gashcrumb:
Tests failed, but I know why, so I think it's good to merge.

e2e/env.js
```diff
@@ -1,27 +0,0 @@
-module.exports = function () {
-
-  // see list of available hooks here
-  // https://github.com/cucumber/cucumber-js/blob/master/docs/support_files/event_handlers.md
-
-  this.setDefaultTimeout(400 * 1000);
```
jludvice: Hi @kahboom, I know it's too late but I need this file :)
Maybe I should rename the file to `env.ts` ?

kahboom: That's because of the `.gitignore` file. I suggest opening up a PR to update that instead, and check the file back in. You can add it as an exception right after `/e2e/*.js`,  like `!/e2e/env.js`.

PR #326, opened 2017-03-28
Fix typo in user's alias

dsimansk:
Just small typo fix in `users.json.example` to link the narrative's Camilla to e2e's Camilla.

cc @jludvice @kahboom


jludvice:
Nice catch @dsimansk, thanks :)

Issue #328, opened 2017-03-28
Create integration fails silently


jludvice:
After creating integration in ipaas-ui and clicking on publish,
I'm redirected on Integration list page, but my integration is **not present** and **I didn't see any error message**.

Inspecting Chrome -> Network tab, I can see HTTP 400 - **bad request** with response
to url https://ipaas-staging.b6ff.rh-idev.openshiftapps.com/api/v1/integrations
with response
```
Unrecognized field &quot;statusType&quot; (class com.redhat.ipaas.model.integration.Integration$Builder), not marked as ignorable
```

![screenshot_20170328_140135](https://cloud.githubusercontent.com/assets/8707241/24403877/289968d8-13bf-11e7-9455-bc3cf434e495.png)

jludvice:
CC @gashcrumb @kahboom @dsimansk

gashcrumb:
Yep, probably need to update the model looks like.  You'll also notice there's a TODO in the code on error handling at the moment :-)

jludvice:
Status update - now i see some error master on commit 28e8950fb2a93468cb99aa7b323ab50ba3b2b970
![integration-error](https://cloud.githubusercontent.com/assets/8707241/24445870/d1cf9e84-146b-11e7-9ba9-5fa8641512f0.png)

gashcrumb:
Yup, that's it *not* failing silently :-)  But we get a timeout error from the backend which we can't do much about in the frontend, so I think this particular issue is all set from the UI perspective.

@sjcox-rh Do we have an 'error saving' design?  I stuck in this quick alert as a stopgap so we at least don't lose the user's work on the frontend side...

KurtStam:
@gashcrumb @jludvice Can list any server exception that reaches the UI here? I can't really see what exception is thrown above exactly since it is a screenshot. We are attempting to handle them all, but it looks like some are slipping through. So if you list them I can go after them. Note that this won't fix the gateway timing out exception. That needs to be address in the whatever the gateway is on OpenShift OnLine.

gashcrumb:
So that's from the gateway timeout, I think we've some code client-side that's seeing that null response and throwing that error possibly.

kahboom:
Was this resolved?

gashcrumb:
As far as the UI is concerned, yes...

kahboom:
Okay, closing this for now then.

PR #329, opened 2017-03-28
Create integration

jludvice:


jludvice:
Found #328 along the way

e2e/integrations/edit/edit.po.ts
```diff
@@ -14,25 +15,22 @@ export class FlowConnection {
    * @returns {webdriver.promise.Promise<boolean>}
    */
   isActive(): P<boolean> {
-    return this.element.element(by.css('p.round.active')).isPresent();
+    return this.element.element(by.css('p.icon.active')).isPresent();
```
kahboom: Sorry! :) But it's more accurate now.

e2e/integrations/edit/edit.po.ts
```diff
@@ -57,6 +67,26 @@ export class ConnectionSelectComponent implements IPaaSComponent {
 }
 
 
+export class IntegrationBasicsComponent implements IPaaSComponent {
+  static readonly nameSelector = 'input[name="nameInput"]';
```
kahboom: I wonder if Protractor supports selecting by the model property instead of by the field property. It's worth looking into. I know they did for Angular 1.x but haven't had a chance to check for 2.x. Either way, this is better than using an ID or class. 👍

jludvice:
Yes, there was support for `by.model` (and still is in protractor 5.x)
http://www.protractortest.org/#/api?view=ProtractorBy.prototype.model

but it's incompatible with Angular 2.x according to https://github.com/angular/protractor#compatibility
and angular 2.x migration guide https://angular.io/docs/ts/latest/guide/upgrade.html#!#appendix-upgrading-phonecat-tests

PR #334, opened 2017-03-28
Add error alert to save or add step page, regenerate model

gashcrumb:
This shows an error if we get one back on the save or add step page and doesn't exit the integration so you at least have a chance to try saving again :-)

Also regenerate the model so we have updated type definitions.

### DIFF
--- a/package.json
+++ b/package.json
@@ -75,7 +75,7 @@
     "sw-precache": "5.0.0",
     "swagger-js-codegen": "1.7.12",
     "ts-node": "2.1.0",
-    "tslint": "4.4.2",
+    "tslint": "^4.5.1",
     "typescript": "2.1.6",
     "webdriver-manager": "12.0.2"
   }
--- a/src/app/app.module.ts
+++ b/src/app/app.module.ts
@@ -120,6 +120,12 @@ export function restangularProviderConfigurer(restangularProvider: any, config: 
     };
   });
 
+  restangularProvider.addErrorInterceptor((response, subject, responseHandler) => {
+    console.log("Response: ", response);
+
+    return true;
+  });
+
   restangularProvider.addResponseInterceptor((data: any, operation: string) => {
     if (operation === 'getList' && data && Array.isArray(data.items)) {
       const pagingData = data.items;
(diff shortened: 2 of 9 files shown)


package.json
```diff
@@ -75,7 +75,7 @@
     "sw-precache": "5.0.0",
     "swagger-js-codegen": "1.7.12",
     "ts-node": "2.1.0",
-    "tslint": "4.4.2",
+    "tslint": "^4.5.1",
```
kahboom: If possible, we should always be using the same patch versions as well. I try to stick to `4.5.1`, for example.

src/app/app.module.ts
```diff
@@ -120,6 +120,12 @@ export function restangularProviderConfigurer(restangularProvider: any, config:
     };
   });
 
+  restangularProvider.addErrorInterceptor((response, subject, responseHandler) => {
+    console.log("Response: ", response);
```
kahboom: I'm guessing Mr. Lint won't be too happy about this? ;D

src/app/model.ts
```diff
@@ -54,25 +54,25 @@ export interface ComponentProperty extends BaseEntity {
...
     organization: Organization;
+    position: string;
     connectorId: string;
     connector: Connector;
-    userId: string;
     description: string;
-    position: string;
-    configuredProperties: {};
+    tags: Array < Tag >
+    ;
```
kahboom: What is this poor, random semicolon doing here? :joy:

src/app/store/entity/entity.store.ts
```diff
@@ -64,6 +64,7 @@ export abstract class AbstractStore<T extends BaseEntity, L extends Array<T>,
         this._loading.next(false);
       },
       (error) => {
+        error = this.massageError(error);
```
kahboom: Wait, am I reading this correctly? Are we massaging errors now? Because I'm the one that really needs the massage here.. :neutral_face:

src/app/store/entity/entity.store.ts
```diff
@@ -118,13 +120,30 @@ export abstract class AbstractStore<T extends BaseEntity, L extends Array<T>,
       });
   }
 
+  private massageError(error: any) {
```
kahboom: Guess so.. 😂

PR #342, opened 2017-03-29
Implement connector validation in the create connection wizard

gashcrumb:


src/app/connections/view/view.component.ts
```diff
@@ -43,11 +46,62 @@ export class ConnectionViewComponent implements OnInit, OnDestroy {
...
+      setTimeout(() => {
+        this.validating = false;
+        let errorHit = false;
+        (<Array<any>>resp).forEach((info) => {
+          if (!errorHit) {
+            if (info['status'] === 'ERROR') {
+              errorHit = true;
+              this.validateError = (<Array<any>>info)['errors'].map((err) => {
+                return err['description'];
+              }).join(', \n');
```
rhuss: Yeah, looks good ;-) For a next iteration we can be a more clever as for missing and invalid options we get the keys directly and so we could lookup the labels for the keys to create a nice and not so technical error message. 

This can be all done on the backend. So the structure of the error values will change to make it easier to consume. Something for the next sprint ;-)

PR #350, opened 2017-03-30
Fix #347 and fix #345

### LINKED ISSUE #345: Create Integration - Hide popup menu in connection list
Like this shouldn't show up.  Also selecting a connection feels wonky, like the click handler doesn't take effect necessarily.

screencapture-localhost-4200-integrations-create-connection-select-0-1490818133673

### LINKED ISSUE #347: Connection List: boxed-area clickable size might be too small
Connections listed in boxes are pretty hard to select when not clicked to `Title` or area around it. User might expect that whole boxed-area is clickable. Same applies for selecting connector in create connection page.

Could we rearrange the div a bit better to enlarge registered area for click to benefit from relatively large boxed-area.

screen shot 2017-03-30 at 11 08 17

gashcrumb:


src/app/connections/list/list.component.html
```diff
@@ -63,39 +63,36 @@ <h4 class="modal-title pull-left">Confirm Delete</h4>
...
-                  <li role="menuitem">
-                    <a class="dropdown-item">Duplicate</a>
-                  </li>
-                  -->
-                  <li>
-                    <a (click)="requestDelete(connection)"
-                       class="dropdown-item">Delete</a>
-                  </li>
-                </ul>
+                  <ul dropdownMenu class="dropdown-menu dropdown-menu-right" role="menu">
```
gashcrumb: @kahboom FYI this is the line that differs from the docs, just no `*` like in the docs for now...  We'll fix these templates then when we update to 1.6.6.  But anyways, we should use this for any dropdowns and not the patternfly jquery code.

kahboom: Right, which is why I was a bit concerned about pushing the upgrade for next sprint. It shouldn't cause any conflicts?

gashcrumb: yeah, I wouldn't make the update now, as there'd also be an impact on the data mapper component, that has a dependency now on ng2-bootstrap as well, so we'd want to ensure we're all aligned.

And who knows, we may also want to consider bumping our angular deps as well to 4.x next week?

PR #357, opened 2017-03-31
Datamapper update

gashcrumb:
So the UI does show up still, but I've hit two issues/question preventing much more from happening:

https://github.com/redhat-ipaas/datamapper-ui/issues/15

and

https://github.com/redhat-ipaas/datamapper-ui/issues/14


src/app/integrations/edit-page/step-configure/data.mapper.example.host.component.ts
```diff
@@ -23,11 +24,11 @@ const MAPPING_KEY = 'atlasmapping';
 @Component({
   selector: 'ipaas-data-mapper-host',
   template: `
-    <div *ngIf="cfg.mappingInputJavaClass && cfg.mappingOutputJavaClass && cfg.mappings">
+    <div *ngIf="cfg.mappings">
```
kahboom: rofl I love this

Issue #93, opened 2016-12-22
Set up CD for E2E Tests


kahboom:
Create a staging environment using CD for the E2E tests that, using GitHub hooks, spin up a Selenium server and, subsequently, run Protractor tests with each push to master for both the `redhat-ipaas/ipaas-client` and `redhat-ipaas/ipaas-client-java` GitHub repositories (so, run the tests for each repository update, irrespective of one another). The staging environment should be on our dedicated OpenShift.

This has already been done for each of the repositories themselves.

kahboom:
Assigning to @dsimansk for now, but whoever gets to it first can change this ;D.

dsimansk:
There's a [small blocker](https://github.com/jludvice/ipaas-ui/commit/36b00cdd662968067952c0a4298ea918a867f12e#commitcomment-21186396), but it seems that solution is near.

kahboom:
@dsimansk @jludvice @iocanel - How is this looking?

jludvice:
Here is my expectations on build pipeline.

I expect that following things will happen in the pipeline/jenkins job - before e2e tests are started
* **deploy ipaas into test openshift** (ipaas-ui, ipaas-rest, keycloak, ...?)
    * build the ipaas-ui from current sources (current PR for instance)
    * this will be time consuming because it will build docker containers and launch pods
* install required tools - browser + xwindow or xvfb (maybe we can have docker image with required tools pre-build?)
* provide GitHub login credentials 
   * either expose as shell variables
   * or inject some config file with jenkins
* provide url to ipaas ui
* launch e2e tests (they will open browser on ipaas-ui deployed in openshift)
* archive cucumber html report
* clean up

Test execution will be the same on developers laptop and in the pipeline.
Tests won't contain any code related to infrastructure/automation.

jludvice:
refactor protractor conf to support multiple envs - #332

PR #369, opened 2017-04-03
Add cmd e2e:xvfb to execute e2e tests in container

dsimansk:
Related to #93. This PR adds possibility to execute e2e tests from Docker container.

### DIFF
--- a/e2e-xvfb.sh
+++ b/e2e-xvfb.sh
@@ -1,0 +1,7 @@
+#!/bin/bash
+
+nohup /usr/bin/Xvfb :99 -ac -screen 0 1280x1024x24 &
+export DISPLAY=:99
+yarn
+npm rebuild node-sass
+exec yarn e2e:ipaas-qe
--- a/e2e-xvfb.sh
+++ b/e2e-xvfb.sh
@@ -3,5 +3,5 @@
 nohup /usr/bin/Xvfb :99 -ac -screen 0 1280x1024x24 &
 export DISPLAY=:99
 yarn
-npm rebuild node-sass
+yarn webdriver-manager update
 exec yarn e2e:ipaas-qe
--- a/e2e/conf/protractor.base.conf.js
+++ b/e2e/conf/protractor.base.conf.js
@@ -18,7 +18,10 @@ exports.config = {
     testPath +  '/**/*.feature'
   ],
   capabilities: {
-    'browserName': 'chrome'
+    'browserName': 'chrome',
+    'chromeOptions': {
+      'args': ['--no-sandbox']
+    }
   },
   directConnect: true,
   baseUrl: 'http://localhost:4200/',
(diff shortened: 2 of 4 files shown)


dsimansk:
To function properly this PR depends on #368.

e2e-xvfb.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+nohup /usr/bin/Xvfb :99 -ac -screen 0 1280x1024x24 &
+export DISPLAY=:99
+yarn
+npm rebuild node-sass
```
jludvice: Do we actually need `npm rebuild node-sass` step?
We don't use local ipaas-ui right?

e2e/conf/protractor.base.conf.js
```diff
@@ -18,7 +18,10 @@ exports.config = {
     testPath +  '/**/*.feature'
   ],
   capabilities: {
-    'browserName': 'chrome'
+    'browserName': 'chrome',
+    'chromeOptions': {
+      'args': ['--no-sandbox']
+    }
```
jludvice: is this option common to all protractor configurations ?
If it's just for `yarn e2e:ipaas-qe`, we might move `chromeOptions` to `e2e/conf/protractor.ipaas-qe.conf.js`

e2e-xvfb.sh
dsimansk: Well, not sure about that. We might actually need `yarn webdriver-manager update`.

e2e/conf/protractor.base.conf.js
dsimansk: I'd keep it common for know, if there aren't any objections :)

e2e-xvfb.sh
```diff
@@ -3,5 +3,5 @@
 nohup /usr/bin/Xvfb :99 -ac -screen 0 1280x1024x24 &
 export DISPLAY=:99
 yarn
-npm rebuild node-sass
+yarn webdriver-manager update
```
jludvice: good catch !

jludvice:
Update: #368 is merged

e2e-xvfb.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+nohup /usr/bin/Xvfb :99 -ac -screen 0 1280x1024x24 &
```
kahboom: Just for reference, the projector for RH Summit has a 16:9 aspect ratio and 1280×720 resolution.

e2e/conf/protractor.base.conf.js
kahboom: I agree with @jludvice , probably best to move to `e2e/conf/protractor.ipaas-qe.conf.js` eventually, but for now is not a big deal at all. 👍

PR #371, opened 2017-04-03
[E2E] fix method to count connections in list

### LINKED ISSUE #370: [E2E] connection count is using wrong element
It seems that following element has changed from `a` to `p`

https://github.com/redhat-ipaas/ipaas-ui/blob/master/e2e/connections/list/list.po.ts#L15

```typescript
const found = await this.rootElement().all(by.css('a.card-title')).getWebElements();
```

```typescript
const found = await this.rootElement().all(by.css('p.card-title')).getWebElements();
```

dsimansk:
This PR should fix #370.


jludvice:
+1, it works fine

PR #382, opened 2017-04-04
Tweak init logic, set parent container height

gashcrumb:
This makes sure the data mapper component has the right height so the left/right panels scroll individually.

### DIFF
--- a/src/app/integrations/edit-page/step-configure/data.mapper.example.host.component.ts
+++ b/src/app/integrations/edit-page/step-configure/data.mapper.example.host.component.ts
@@ -31,11 +31,18 @@ const MAPPING_KEY = 'atlasmapping';
 @Component({
   selector: 'ipaas-data-mapper-host',
   template: `
-    <div *ngIf="initialized">
+    <div *ngIf="initialized" class="data-mapper-host">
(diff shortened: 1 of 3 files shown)


gashcrumb:
lemme also push a commit to update the datamapper component before merging this

gashcrumb:
@kahboom k, ready to merge!

src/app/integrations/edit-page/step-configure/data.mapper.example.host.component.ts
```diff
@@ -135,27 +141,35 @@ export class DataMapperHostComponent extends FlowPage implements OnInit, OnDestr
...
+          this.initializeMapper();
+        }, 10);
       },
     });
   }
 
   initializeMapper() {
+    this.cfg.mappingService.mappingUpdated$.subscribe(() => {
+      this.detector.detectChanges();
+    });
```
kahboom: This lack of spaces between functions is driving me nuts. :see_no_evil:

kahboom: If it bothers you too, we should consider updating the lint config.. although I do agree with forcing a new line at the end of each file.

Issue #376, opened 2017-04-03
Working data mapper component


gashcrumb:
Data mapper component is integrated, but doesn't yet work.  Needs to fetch a pom for the current integration the user is working on so that it can supply it to the data mapper backend.

chirino:
Rest API now has the following endpoint: /api/v1/integration-support/generate/pom.xml
POST to it an the current integration.  The integration does not need to exist yet.

gashcrumb:
@chirino I now have the data mapper sending along that pom.xml file with the linked PR.  The mapper component is getting back this response when it goes to use it:

```
{
  "MavenClasspathResponse": {
    "jsonType": "com.mediadriver.atlas.java.v2.MavenClasspathResponse",
    "classpath": "Command returned non-zero exit code: 1",
    "executionTime": 3009
  }
}
```

not really sure what that could mean :-)  Anyways, once the PR is merged it'll be available in staging, I've also created an integration with a mapper step, aptly titled `Test With Mapping`.

kahboom:
Merged your PR @gashcrumb .

chirino:
I think I know what that means.. It can't find mvn.  I'll see if I can fix the docker image.

gashcrumb:
Almost there, hit this on staging -> https://github.com/redhat-ipaas/datamapper-ui/issues/23

gashcrumb:
This PR will also affect this issue -> https://github.com/redhat-ipaas/ipaas-rest/pull/276

gashcrumb:
Also hitting this now -> https://github.com/redhat-ipaas/datamapper-ui/issues/24

gashcrumb:
It works so I think this issue is all set at least.

PR #391, opened 2017-04-05
Consistent save/load of datamapper data

gashcrumb:
For #376


src/app/integrations/edit-page/step-configure/data.mapper.example.host.component.ts
```diff
@@ -116,13 +112,18 @@ export class DataMapperHostComponent extends FlowPage implements OnInit, OnDestr
     if (mappings) {
       const mappingDefinition = new MappingDefinition();
       // Existing mappings, load from the route
+      try {
       this.mappingService.deserializeMappingServiceJSON(mappings, mappingDefinition);
+      } catch (err) {
+        // TODO popup or error alert?  At least catch this so we initialize
```
kahboom: Definitely, we'll need that global notification system in place or at least an inline notification.

PR #381, opened 2017-04-04
Create connection test

jludvice:
Not ready for merge - opening for discussion about implementation.

PR #399, opened 2017-04-06
[jquery types] Fix type errors introduced by including @types/jqueryi…

jludvice:
…n #381


jludvice:
hmm, too late, jquery typings were already removed in #398

gashcrumb:
Ah yeah, you'll want to rebase and then add that in as a `devDependency`

jludvice:
rebase done and tests just passed

PR #445, opened 2017-04-13
Add an example config.json that should work for testing against a local minishift install.

chirino:


jimmidyson:
Bit confused how this differs to `config.json.example`? They're just different values, right? But ones that certainly don't work for me... IP address is nothing like mine...

PR #446, opened 2017-04-13
Add more icons

jimmidyson:
Fixes https://issues.jboss.org/browse/IPAAS-295


jimmidyson:
Hope I've resized these properly, kept transparent backgrounds as much as possible too so should look ok. Let me know if they should have white backgrounds or whatever.

PR #451, opened 2017-04-14
Make the dashboard’s Top 5 Integrations card responsive.

chirino:


chirino:
before:
red_hat_ipaas
after:
red_hat_ipaas

src/app/dashboard/integrations.component.scss
```diff
@@ -216,17 +187,19 @@
...
             }
-
             .fa.fa-angle-right {
               font-size: 18px;
             }
           }
         }
 
         .name {
+          margin-left: 1em;
```
kahboom: For now this is fine but would be good to use a mobile framework post-Summit.

PR #465, opened 2017-04-17
IPAAS-301: Add an auto-link-github option to enable/disable doing the GitHub login.

chirino:


gashcrumb:
Does this require folks to update their local copies of config.json?  Or will it just continue working the same?

chirino:
Yeah, it's backward compatible with previous config.json
So no need to update.

PR #469, opened 2017-04-18
Use the right icons in the integration editor

gashcrumb:
screencapture-localhost-4200-integrations-edit-1-save-or-add-step-1492526418119

### DIFF
--- a/src/app/integrations/edit-page/flow-view/flow-view-step.component.html
+++ b/src/app/integrations/edit-page/flow-view/flow-view-step.component.html
@@ -50,12 +50,7 @@
        (mouseover)="showTooltip()"
        (mouseleave)="hideTooltip()"
        container="body">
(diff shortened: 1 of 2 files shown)


src/app/integrations/edit-page/flow-view/flow-view-step.component.html
```diff
@@ -51,8 +51,11 @@ <h4 class="modal-title pull-left">Confirm Delete</h4>
        (mouseleave)="hideTooltip()"
        container="body">
       <!-- Only display icon if step kind is not an endpoint or log -->
+      <img class="connection-icon" *ngIf="step.stepKind === 'endpoint'" src="../../../assets/icons/{{ step.connection.connectorId }}.integration.png">
+      <!--
```
jimmidyson: Worth deleting this rather than commenting out?

PR #470, opened 2017-04-18
Use error icon in the integration list if we get a status message back, also ensure the integration editor works again

gashcrumb:


src/app/integrations/edit-page/flow-view/flow-view-step.component.html
```diff
@@ -50,7 +50,11 @@ <h4 class="modal-title pull-left">Confirm Delete</h4>
        (mouseover)="showTooltip()"
        (mouseleave)="hideTooltip()"
        container="body">
-      <img class="connection-icon" *ngIf="step.stepKind === 'endpoint'" src="../../../assets/icons/{{ step.connection.connectorId }}.integration.png">
+      <i [class]="getIconClass()"
+         *ngIf="step.stepKind === 'endpoint' && !step.connection"></i>
+      <img class="connection-icon"
+           *ngIf="step.stepKind === 'endpoint' && step.connection"
```
rhuss: guess this is the fix ;-)

PR #480, opened 2017-04-19
Add tabs to integration list for filtering state

gashcrumb:
This adds the tabs that filter the integration list by state, also I fixed how the end icon is shown while sorting merge conflicts.

screencapture-localhost-4200-integrations-1492620798155


src/app/common/object-property-filter.pipe.ts
```diff
@@ -27,6 +27,10 @@ export class ObjectPropertyFilterPipe {
     }
     return objects.filter((obj: any) => {
       const value = getPropertyValue(obj, config.propertyName);
+      if (value === undefined) {
```
kahboom: stop, drop, and roll

PR #494, opened 2017-04-20
Load e2e test creadentials from env vars

dsimansk:
This PR should add option to specify credential via env vars.

Cc @iocanel, sorry took a bit longer due to my PTO.

### DIFF
--- a/e2e/common/common.ts
+++ b/e2e/common/common.ts
@@ -12,8 +12,16 @@ export class User {
     this.alias = alias;
     this.description = description;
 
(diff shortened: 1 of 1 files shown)


dsimansk:
Env variable to use:
```bash
IPAAS_<user_alias>_USERNAME
IPAAS_<user_alias>_PASSWORD
```

The alias is used to determine role. We currently use alias `Camilla` per narratives as default user. In future we should have at least 2 (maybe 2+) different roles.

PR #496, opened 2017-04-20
Test to verify Dashboard links

lvydra:
Test to verify links of Dashboard work correctly and point to desired destinations.


gashcrumb:
@lvydra there's lint errors to fix, run `yarn lint` locally to see them.

e2e/common/common.steps.ts
```diff
@@ -86,6 +102,27 @@ class CommonSteps {
       .to.eventually.be.true.notify(callback);
   }
 
+  @then(/^she is presented with the "([^"]*)" elements*$/)
+  public expectElementsPresent(elementClassNames: string, callback: CallbackStepDefinition): void {
+
+    const elementClassNamesArray = elementClassNames.split(',');
```
jludvice: Can we find another way of implementing this step?

```gherkin
# I really like this lavel ob abstraction - it seems pretty readable
When Camilla clicks on the "View All Integrations" link

# defining dom elements or css classes should remain in page objects imho
Then she is presented with the "top-integrations,integration-board,recent-updates,ipaas-dashboard-connections" elements
```

What do you think @lvydra @dsimansk  ?

e2e/common/common.steps.ts
```diff
@@ -75,6 +83,14 @@ class CommonSteps {
       .catch((e) => callback(e));
   }
 
+  @when(/clicks? on the "([^"]*)" link.*$/)
+  public clickOnLink(linkTitle: string, callback: CallbackStepDefinition): void {
+    this.world.app.clickLink(linkTitle)
+      .then(() => callback())
+      // it may fail but we still want to let tests continue
+      .catch((e) => callback(e));
+  }
```
jludvice: this note applies also to myself ;)

we may use either callback
```ts
  @when(/clicks? on the "([^"]*)" link.*$/)
  public clickOnLink(linkTitle: string, callback: CallbackStepDefinition): void {
    this.world.app.clickLink(linkTitle)
      .then(() => callback())
      // it may fail but we still want to let tests continue
      .catch((e) => callback(e));
  }
```
or return promise (instead of callback method attribute we define Promise as return value)
- specifyng both callback attribute and also promise return value throws error
- in this case data passed to `Promise.reject` (or `P.reject`) is displayed in test report as error message.
```ts
  @when(/^type "([^"]*)" into connection description/)
  public typeConnectionDescription(description: string): P<void> {
    // Write code here that turns the phrase above into concrete actions
    const connectionView = new ConnectionViewComponent();
    return connectionView.description.set(description);
  }
```

e2e/common/common.steps.ts
dsimansk: @lvydra I would be better to look up the table by title like `Integration Board`, `Recent Updates` etc. to avoid specific element declaration in steps file. Could you please take a look?

jludvice: btw there is lookup ready for us by protractor: `by.cssContainingText()`
Exmaple: https://github.com/redhat-ipaas/ipaas-ui/blob/afae5ceb2d3849f55528e03f24387d6704eb519c/e2e/integrations/edit/edit.po.ts#L52

PR #504, opened 2017-04-20
Integration List

kahboom:
screenshot 2017-04-20 13 34 11

Changes:
- Fix icon padding.
- Center Draft label.
- Add toggle for Active/Inactive.
- Adjust row distribution of elements.

NOTE: Toggle is not reflecting actual Active/Inactive setting. The value is obtained but this is an issue related to the toggle we are using. Fix in next PR.

### DIFF
--- a/src/app/integrations/list/list.component.html
+++ b/src/app/integrations/list/list.component.html
@@ -104,13 +104,17 @@
               </h3>
             </div>
 
-            <div *ngIf="integration.currentStatus !== 'Pending'">
+            <div *ngIf="integration.currentStatus !== ('Pending' || 'Draft')">
               <label class="switch_wrap" for="field-{{ integration.id }}">
                 <input id="field-{{ integration.id }}"
                        type="checkbox"
                        value="1" />
                 <div class="switch"><span class="bullet"></span></div>
               </label>
+            </div>
+
+            <div *ngIf="integration.currentStatus === 'Draft'">
+              <span class="label label-warning" disabled>Draft</span>
             </div>
 
             <!--
--- a/src/app/integrations/list/list.component.html
+++ b/src/app/integrations/list/list.component.html
@@ -107,10 +107,11 @@
             </div>
 
             <div *ngIf="integration.currentStatus !== 'Pending' && integration.currentStatus !== 'Draft'">
-              <label class="switch_wrap" for="field-{{ integration.id }}">
+              <label class="switch_wrap"
+                     for="field-{{ integration.id }}">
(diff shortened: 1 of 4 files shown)


src/app/integrations/list/list.component.scss
```diff
@@ -85,9 +102,130 @@
...
+            .label {
+              cursor: pointer;
+              padding: 6px;
+            }
+          }
+
+
+          // Other States - Toggle Button
+
+          .switch_wrap input[type="checkbox"] {
```
kahboom: We shouldn't be using these either, but this was literally a copy and paste from an example. :/ *runs and hides*

src/app/integrations/list/list.component.html
```diff
@@ -106,27 +106,28 @@ <h4 class="modal-title pull-left">Confirm {{ getActionTitle() }}</h4>
...
             </div>
           </div>
         </div>
 
 
         <!-- Metrics & Actions -->
 
-        <div class="col-xs-7">
+        <div class="col-xs-7"
+             (click)="goto(integration)">
```
gashcrumb: I'd leave out this click handler...

kahboom: Okay will do.

kahboom: Leaving it for the rest of the sub-divisions though like the `uses` etc. I forgot the kabob was in that same group. :/

src/app/integrations/list/list.component.html
```diff
@@ -106,27 +106,28 @@ <h4 class="modal-title pull-left">Confirm {{ getActionTitle() }}</h4>
...
-              <span *ngIf="integration.currentStatus === 'Deactivated'"
-                    class="label label-default"
-                    (click)="requestActivate(integration)">Inactive</span>
-              <span *ngIf="integration.currentStatus === 'Draft'"
-                    class="label label-warning"
-                    disabled>Draft</span>
-              <span *ngIf="integration.currentStatus === 'Deleted'"
-                    class="label label-danger"
-                    disabled>Deleted</span>
+            <div *ngIf="integration.currentStatus !== 'Pending' && integration.currentStatus !== 'Draft'">
```
gashcrumb: maybe just check if it's `===` `'Activated'` or `'Deactivated'` instead?  Otherwise this might show up in the `Deleted` case?

kahboom: Yeah, good call.

kahboom:
screenshot 2017-04-20 13 48 05

PR #516, opened 2017-04-24
Rm Toggle, Add Labels

kahboom:
## Changes

- Remove the toggle and ability to change status from status label.
- Re-add labels.
- Hide 'Uses' column.
- Fix actions in dropdown kebab based on `currentStatus`.

## Screenshots

screenshot 2017-04-24 13 16 10

NOTE: We'll need to re-assess the conditions for actions using a diagram. I'll work on that throughout the week.

### DIFF
--- a/src/app/integrations/list/list.component.html
+++ b/src/app/integrations/list/list.component.html
@@ -109,7 +109,7 @@
           <div *ngIf="integration.currentStatus !== 'Pending'"
                class="not-pending">
               <span class="label label-{{ getLabelClass(integration.currentStatus) }}"
-                    disabled>{{ integration.currentStatus }}</span>
+                    disabled>{{ getStatusText(integration.currentStatus) }}</span>
           </div>
         </div>
 
@@ -176,7 +176,7 @@
                 </li>
                 <li role="menuitem"
                     *ngIf="canDeactivate(integration.currentStatus)">
-                  <a (click)="requestActivate(integration)">Deactivate</a>
+                  <a (click)="requestDeactivate(integration)">Deactivate</a>
                 </li>
                 <!--
                     <li>
--- a/src/app/integrations/list/list.component.html
+++ b/src/app/integrations/list/list.component.html
@@ -93,60 +93,44 @@
 
         <!-- Active/Inactive Labels -->
 
-        <div class="col-xs-5 active-inactive">
-          <div class="list-view-pf-additional-info-item">
+        <div class="col-xs-6 active-inactive">
+          <!-- In Progress -->
 
-            <!-- In Progress -->
+          <div class="in-progress"
+               *ngIf="integration.currentStatus === 'Pending'">
+            <h3>
+              <div class="spinner spinner-sm spinner-inline"></div>
+              In Progress
+            </h3>
+          </div>
 
-            <div class="in-progress"
-                 *ngIf="integration.currentStatus === 'Pending'">
-              <h3>
-                <div class="spinner spinner-sm spinner-inline"></div>
-                In Progress
-              </h3>
-            </div>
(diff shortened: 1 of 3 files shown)


src/app/integrations/list/list.component.html
```diff
@@ -93,60 +93,44 @@ <h4 class="modal-title pull-left">Confirm {{ getActionTitle() }}</h4>
...
+          <!-- Status -->
 
-            <div *ngIf="integration.currentStatus === 'Deleted'"
-                 class="deleted">
-              <span class="label label-danger" disabled>Deleted</span>
-            </div>
+          <div *ngIf="integration.currentStatus !== 'Pending'"
+               class="not-pending">
+              <span class="label label-{{ getLabelClass(integration.currentStatus) }}"
+                    disabled>{{ integration.currentStatus }}</span>
```
gashcrumb: We need to show 'Active' or 'Inactive', but using this unfortunately will mean we'll show 'Activated' and 'Deactivated' which is inconsistent with the wording on the dashboard.

Only reason I mention this is because I fixed this issue -> https://github.com/redhat-ipaas/ipaas-ui/issues/487

src/app/integrations/list/list.component.html
```diff
@@ -182,28 +166,25 @@ <h4 class="modal-title pull-left">Confirm {{ getActionTitle() }}</h4>
...
-                    *ngIf="integration.desiredStatus === 'Deactivated'">
+                    *ngIf="canActivate(integration.currentStatus)">
                   <a (click)="requestActivate(integration)">Activate</a>
                 </li>
                 <li role="menuitem"
-                    *ngIf="integration.active === false"
-                    (click)="requestActivate(integration)">
-                  <a>Activate</a>
+                    *ngIf="canDeactivate(integration.currentStatus)">
+                  <a (click)="requestActivate(integration)">Deactivate</a>
```
gashcrumb: Shouldn't this be `requestDeactivate`?

src/app/integrations/list/list.component.ts
```diff
@@ -160,6 +151,69 @@ export class IntegrationsListComponent {
...
+        return false;
+      default:
+        return true;
+    }
+  }
+
+  canEdit(currentStatus) {
+    switch (currentStatus) {
+      case 'Deleted':
+      case 'Draft':
```
gashcrumb: I think a `Draft` integration can be edited

src/app/integrations/list/list.component.ts
```diff
@@ -160,6 +151,69 @@ export class IntegrationsListComponent {
...
+      case 'Pending':
```
gashcrumb: Also no reason you can't edit a `Pending` one.

src/app/integrations/list/list.component.html
kahboom: @gashcrumb Good catch, I didn't add that in this PR but never noticed it!

src/app/integrations/list/list.component.ts
kahboom: +1

src/app/integrations/list/list.component.ts
kahboom: k

src/app/integrations/list/list.component.html
kahboom: Np!

src/app/integrations/list/list.component.ts
```diff
@@ -242,6 +240,19 @@ export class IntegrationsListComponent {
     }
   }
 
+  getStatusText(currentStatus) {
```
gashcrumb: +1

PR #536, opened 2017-05-09
e2e: Refactor login flow

dsimansk:
@kahboom @jludvice 

I've happened to redo a login flow in e2e tests again, inherently getting out all async nature of it :). The previous version wasn't very stable for CI to run it properly. Could you please take a look? There might be few more touch ups needed.

### DIFF
--- a/e2e/app.po.ts
+++ b/e2e/app.po.ts
@@ -130,24 +130,27 @@ export class AppPage {
     // need to disable angular wait before check for current url because we're being redirected outside of angular
     browser.waitForAngularEnabled(false);
 
-    this.goToUrl(AppPage.baseurl);
+    await this.goToUrl(AppPage.baseurl);
 
-    // wait either for login page or loaded ipaas app
-    await P.race([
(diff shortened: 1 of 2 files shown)


e2e/app.po.ts
```diff
@@ -130,24 +130,27 @@ export class AppPage {
...
-      browser.wait(ExpectedConditions.presenceOf(element(by.css('input#firstName'))), 1000, 'Keycloack login page' ).then(() => {
-        log.info('Keycloak first time visit page');
-        return new KeycloakDetails().submitUserDetails(user.userDetails);
-      }),
-    ]);
+    await this.goToUrl(AppPage.baseurl);
+
+    let currentUrl = await browser.getCurrentUrl();
+    if (contains(currentUrl, 'github.com/login')) {
+      log.info('GitHub login page');
```
kahboom: Wonder if there is a better way to document and log this, possibly using the testing framework.

jludvice: what do you mean @kahboom ?

PR #539, opened 2017-05-11
e2e: allign names and bindings

dsimansk:
This PR fixes def articles preceding `Syndesis` in some of the feature->step bindings, and protractor conf file name.


jimmidyson:
Thanks!

PR #542, opened 2017-05-15
Tests to verify delete action

lvydra:
Tests for verify delete action on connections list, integrations list and dashboard


dsimansk:
@kahboom this PR is a bit messy, but could you pls review it?

e2e/connections/list/list.po.ts
```diff
@@ -3,8 +3,6 @@ import { Promise as P } from 'es6-promise';
 import { element, by, ElementFinder } from 'protractor';
 import { SyndesisComponent } from '../../common/common';
 import { log } from '../../../src/app/logging';
```
kahboom: Thank you for using the internal logger.

Issue #550, opened 2017-05-18
Connections List: Kebab menu isn't displayed, but navigates directly to details


dsimansk:
After clicking on Kebab menu on any listed Connections, the Kebab menu is not displayed. But the page navigates directly to Connection detail page.

dsimansk:
Verified commit 09150e0. Kebab menu is diplayed properly with action `{Delete, View}`. 

@gashcrumb one other thought came to my mind. Wouldn't it be better to have `{Delete, View, Edit}` in the menu?

gashcrumb:
Yeah, 'editing' was hacked in quickly though and I think might take a bit of work to support linking to it directly.  Can you raise a new issue for adding 'Edit' into that menu so we can close this one?

dsimansk:
Closing this one and opened https://github.com/syndesisio/syndesis-ui/issues/613 for edit action tracking.

PR #551, opened 2017-05-18
fix: Make kebab menu show up properly for #550

gashcrumb:


src/app/connections/list/list.component.html
```diff
@@ -62,42 +62,41 @@ <h4 class="modal-title pull-left">Confirm Delete</h4>
...
                 </button>
-                <ul *dropdownMenu
-                    class="dropdown-menu dropdown-menu-right"
-                    role="menu">
-                  <li role="menuitem">
-                    <a [routerLink]="['/connections', connection.id]"
-                       class="dropdown-item">View</a>
-                  </li>
-                  <!--
+                  <ul *dropdownMenu
```
kahboom: Sorry, I'm sure we keep going back and forth on this without doing it on purpose lol. We should probably share IDE-specific settings & general style guides.

Issue #549, opened 2017-05-18
Integrations List: Kebab menu is displayed empty for Deleted state integration


dsimansk:
Picture worth a 1000 words :) There're no actions listed for `Deleted` state integration.

![screen shot 2017-05-18 at 13 43 00](https://cloud.githubusercontent.com/assets/5637792/26200742/ef43c6da-3bcf-11e7-8e02-35f4aea5a1ae.png)

gashcrumb:
Probably shouldn't show a menu at all if there's no actions for a given state.  Not sure if 'deleted' as a state will be there going forward.

dsimansk:
You wouldn't guess that I've had a delicious Kebab for lunch today (for real :)). Thanks for quick fixes 👍

kahboom:
lol

dsimansk:
Verified commit b20d068 implements correct behaviour for `Deleted` state.

Issue #548, opened 2017-05-18
Integrations List: Kebab menus open after double click


dsimansk:
First click on Kebab menu highlights square around the menu, but doesn't open the actions itself. The menu listing is opened after second click.

![screen shot 2017-05-18 at 13 40 51](https://cloud.githubusercontent.com/assets/5637792/26200699/b86acca8-3bcf-11e7-8f97-1a3b449d55d0.png)

dsimansk:
Verified commit b20d068 fixed the issue.

PR #552, opened 2017-05-18
fix: Remove data-toggle attribute for #548, hide kebab for #549

gashcrumb:


src/app/integrations/list/list.component.html
```diff
@@ -108,8 +108,8 @@ <h4 class="modal-title pull-left">Confirm {{ getActionTitle() }}</h4>
 
           <div *ngIf="integration.currentStatus !== 'Pending'"
                class="not-pending">
-              <span class="label label-{{ getLabelClass(integration.currentStatus) }}"
-                    disabled>{{ getStatusText(integration.currentStatus) }}</span>
+            <span class="label label-{{ getLabelClass(integration.currentStatus) }}"
```
kahboom: Gah! Those formatting changes. Not your fault, we just need to get on the same page. I don't mind either way but maybe we could throw it in the community site.

PR #592, opened 2017-07-03
Update model to use correct pluralization rules and update some deps for ng4 compatibility

jimmidyson:


src/app/connections/view/view.component.html
```diff
@@ -1,11 +1,8 @@
-<div class="row row-cards-pf connections-view"
-     [ngClass]="{edit: 'edit', create: 'create'}[mode] || 'view'"
-     *ngIf="connection">
+<div class="row row-cards-pf connections-view" [ngClass]="{edit: 'edit', create: 'create'}[mode] || 'view'" *ngIf="connection">
```
gashcrumb: Are we not doing newlines between element attributes anymore?

jimmidyson: Sorry! Damn editor formatting aggressively... Reverted.

jimmidyson:
@gashcrumb Any chance of another look so we can merge?

gashcrumb:
Yep, sorry yesterday was a holiday here, looking now.

PR #611, opened 2017-07-14
Add configuration for data mapper service call debug logging and mock…

jbakermd:
… source/target additions.


jbakermd:
This caused a build error. I need to update the data mapper dependency to point to the latest commit hash with the api changes.

PR #618, opened 2017-07-19
Form field level help

### LINKED ISSUE #585: When creating connection only part of field help messages are visible
When creating connection, the help message for each field is shown directly in the field, which helps a lot. However when the description (the help message) of the field is long, only first part is displayed and _there is no way how to display next part_.

For example, when creating new salesforce connection you will see something like this:

salesforce-connection

For example, when I am not sure what should I fill for refresh token I see only part of message:

> Refresh token already obtained in the refresh token OAuth flow. One needs to setup a web application and co

But whole message is:

> Refresh token already obtained in the refresh token OAuth flow. One needs to setup a web application and configure a callback URL to receive the refresh token or configure using the builtin callback at https://login.salesforce.com/services/oauth2/success or https://test.salesforce.com/services/oauth2/success and then retrive the refresh_token from the URL at the end of the flow. Note that in development organizations Salesforce allows hosting the callback web application at localhost.

Since there is no tooltip nor animation of the text which would display the other parts. There is no way how to display whole help message.

riccardo-forina:
Looking at Patternfly, form field level help should be presented like this: http://www.patternfly.org/pattern-library/forms-and-controls/field-level-help/

The only way to customize the look&feel of form elements with ng2-dynamic-forms I found, is to roll our own UI module which is basically a fork of the default ui-bootstrap one.  

Since this module is gonna be shared across the whole application, it's in the common folder. To use it you have to import `DynamicFormsPatternflyUIModule` into your module instead of the `DynamicFormsBootstrapUIModule` one, and in the templates you have to use the new directive `dynamic-form-patternfly-control` instead of `dynamic-form-bootstrap-control`.  

To show this help icon all you need to do is to add the `hint` property to the field model.  This gets rid of the previous way of showing hints, which was the text under the field like here https://github.com/syndesisio/syndesis-ui/issues/585#issuecomment-316146717.

fixes #585, refs #585

help-icon
tooltip

### DIFF
--- a/src/app/common/forms.service.ts
+++ b/src/app/common/forms.service.ts
@@ -71,8 +71,9 @@ export class FormFactoryService {
         formField = new DynamicTextAreaModel(
           {
             id: key,
-            label: (value.displayName || key) + `<span class="glyphicon glyphicon-info-sign"></span>`,
+            label: value.displayName || key,
             value: value.value || value.defaultValue,
+            hint: value.description,
             required: value.required,
             rows: value.rows,
             cols: value.cols,
@@ -94,9 +95,10 @@ export class FormFactoryService {
         formField = new DynamicInputModel(
           {
             id: key,
-            label: type === 'hidden' ? null : (value.displayName || key) + `<span class="glyphicon glyphicon-info-sign"></span>`,
+            label: type === 'hidden' ? null : value.displayName || key,
             inputType: type,
             value: value.value || value.defaultValue,
+            hint: value.description,
             required: value.required,
           },
           {
--- a/src/app/common/forms.service.ts
+++ b/src/app/common/forms.service.ts
@@ -71,8 +71,7 @@ export class FormFactoryService {
         formField = new DynamicTextAreaModel(
           {
             id: key,
-            label: value.displayName || key,
-            hint: value.description,
+            label: (value.displayName || key) + `<span class="glyphicon glyphicon-info-sign"></span>`,
             value: value.value || value.defaultValue,
             required: value.required,
             rows: value.rows,
@@ -95,8 +94,7 @@ export class FormFactoryService {
         formField = new DynamicInputModel(
           {
             id: key,
-            label: type === 'hidden' ? null : value.displayName || key,
-            hint: type === 'hidden' ? null : value.description,
+            label: type === 'hidden' ? null : (value.displayName || key) + `<span class="glyphicon glyphicon-info-sign"></span>`,
             inputType: type,
             value: value.value || value.defaultValue,
             required: value.required,
--- a/src/app/common/ui-patternfly/dynamic-form-patternfly.component.html
+++ b/src/app/common/ui-patternfly/dynamic-form-patternfly.component.html
@@ -1,0 +1,234 @@
+<div [class.form-group]="asBootstrapFormGroup || model.cls.element.container.includes('form-group')"
+     [class.has-error]="hasErrorMessaging && !hasFocus && isInvalid"
+     [formGroup]="group"
+     [ngClass]="[model.cls.element.container, model.cls.grid.container]">
+
+    <label *ngIf="type !== 2 && model.label !== null"
+           [attr.for]="model.id"
+           [ngClass]="[model.cls.element.label, model.cls.grid.label]">
+      <span class="label-text"
+            [innerHTML]="model.label"></span>
+      <div *ngIf="model.hint !== null"
+           [tooltip]="model.hint"
+           placement="right"
+           class="hint-icon">
+        <span class="glyphicon glyphicon-info-sign"></span>
+      </div>
+    </label>
+
+
+    <ng-template *ngIf="type !== 1 && template?.align === 'START'"
+                 [ngTemplateOutlet]="template.templateRef"
+                 [ngTemplateOutletContext]="model"></ng-template>
+
+
+    <div *ngIf="type === 2" [ngClass]="model.cls.grid.control">
+
+        <div class="checkbox" [class.disabled]="model.disabled">
+
+            <label>
+
+                <input type="checkbox"
+                       [attr.tabindex]="model.tabIndex"
+                       [checked]="model.checked"
+                       [dynamicId]="bindId && model.id"
+                       [formControlName]="model.id"
+                       [name]="model.name"
+                       [ngClass]="model.cls.element.control"
+                       [required]="model.required"
+                       [value]="model.value"
+                       (blur)="onFocusChange($event)"
+                       (change)="onValueChange($event)"
+                       (focus)="onFocusChange($event)"/><span [innerHTML]="model.label"
+                                                              [ngClass]="[model.cls.element.label, model.cls.grid.label]"></span></label>
+        </div>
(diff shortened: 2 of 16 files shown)


riccardo-forina:
Uhm, not sure about the error on CircleCI, if I launch the same command I get no errors.
The only noticeable difference is yarn version, `v0.23.2` on CI vs `v0.27.5` locally.

```yarn ng build -- --aot --prod --progress=false
yarn ng v0.27.5
$ "/Users/riccardoforina/work/RH/syndesis-ui/node_modules/.bin/ng" "build" "--aot" "--prod" "--progress=false"
Template parse warnings:
The <template> element is deprecated. Use <ng-template> instead ("
[WARNING ->]<template [ngIf]="!isClosed">
  <div [class]="'alert alert-' + type" role="alert" [ngClass]="classes""): ng:///Users/riccardoforina/work/RH/syndesis-ui/node_modules/ngx-bootstrap/alert/alert.component.d.ts.AlertComponent.html@1:0
Hash: e3d0a612844c152901f8
Time: 81159ms
chunk    {0} 0.42212758e59e30dff46a.chunk.js 1 MB {1} {2} {3} {5} [rendered]
chunk    {1} 1.c3444a37e9fb511bbeea.chunk.js 788 kB {0} {2} {3} {5} [rendered]
chunk    {2} 2.63985584a114d0124af4.chunk.js 544 kB {0} {1} {3} {5} [rendered]
chunk    {3} 3.ad8188670374244cd5d9.chunk.js 48.6 kB {0} {1} {2} {5} [rendered]
chunk    {4} polyfills.aacc3afe890ed9655444.bundle.js (polyfills) 241 kB {9} [initial] [rendered]
chunk    {5} main.66fb504889dcaa45242a.bundle.js (main) 860 kB {8} [initial] [rendered]
chunk    {6} scripts.d4f5105a172230bd6102.bundle.js (scripts) 914 kB {9} [initial] [rendered]
chunk    {7} styles.b52888d9959243387b0b.bundle.css (styles) 281 bytes {9} [initial] [rendered]
chunk    {8} vendor.39bfac7ac33f846a6543.bundle.js (vendor) 4.06 MB [initial] [rendered]
chunk    {9} inline.2fb7c78271259cc7f28d.bundle.js (inline) 0 bytes [entry] [rendered]
Done in 86.49s.```

**Comment by jimmidyson** [comment]
I get the same error locally with yarn 0.27.5 btw.

**Comment by jimmidyson** [comment]
Let me push a quick fix to this PR branch.

**Comment by jimmidyson** [comment]
BTW @riccardo-forina - thank you so much for the PR! Really appreciated!

**Comment by riccardo-forina** [comment]
Uhm. I took good care to use v1.4.17 of ng2-dynamic-forms because that was the version used at the time of the fork, which had no DynamicFormValidationService. I see now that the project has been updated to 1.4.20, so I guess that's the reason why we are getting failures on the CI env.

**Comment by jimmidyson** [comment]
Oh sorry yeah we had to upgrade the version. Now there's a test error on circle which isn't happening for me locally... If you have time to look at it, that would be appreciated, if not then I'm sure @kahboom or @gashcrumb will be able to help out later hopefully.

**Comment by riccardo-forina** [comment]
Sure thing, I can give it some time later this evening. :thumbsup:

**Comment by riccardo-forina** [comment]
To summarise what happened, before the PR I rebased from upstream as it's normal. 
On upstream the package.json has been updated, but I got no help from the tools to notice this. 
For this reason, I did not launch a `yarn install` before launching the tests again.
Because of this, I locally had version 1.4.17 of ng2-dynamic-forms, which allowed tests to pass.

Had I done updated my modules, I would have got the signature errors and noticed the problem before submitting the PR.
 
A way to avoid problems like this in the future might be a git-hook when pulling that checks for changes on the package.json and gives a _visible_ feedback about this. Something along the lines of this: https://davidwalsh.name/git-hook-npm-install-package-json-modified

**Comment by kahboom** [comment]
@riccardo-forina - Appreciate the PR. With regard to the  ng2-dynamic-forms fork of the ui-bootstrap module, we had not done this previously as didn't want to reinvent the wheel considering PatternFly is already working on an Angular 2 project [here](https://github.com/patternfly/patternfly-ng). Moreover, this would be specific to dynamic forms, and we wanted a solution that would allow us to use PF across all components.

Given time constraints we had for Red Hat Summit (and still have for preview), and that it was not yet available at that point, we had to move forward and using custom styling per component in a hierarchical manner. In the near future, we'd like to refactor this and possibly switch to use PF's new project, but it's just not at the top of the priority list at the moment.

100% with you on that last point of using a git-hook when pulling.

**Comment by gashcrumb** [comment]
@kahboom, I think I'm onboard with this approach in general, as this would save us a lot of hackery and custom templates that we'd have to apply in the forms service anyways.  I'm meh on having a tooltip only because it feels like a lot of text, but that's something that'll be easier to change (if we want to) using this approach.

The only drawback to this approach that I see is exposed here is that we have to be careful to maintain this more closely, so API changes in the upstream will have a greater effect.  In the case here the form array group API changed between 1.4.17 and 1.4.20, the `createGroup` function was renamed to `groupFactory` in this commit -> https://github.com/udos86/ng2-dynamic-forms/commit/f0bb2d433577b4e31747a6c70a3aff9074d580d9

@riccardo-forina So hopefully it's just a matter of doing that rename, I can take a look a bit later if you don't get a chance to.

**Comment by gashcrumb** [comment]
> A way to avoid problems like this in the future might be a git-hook when pulling that checks for changes on the package.json and gives a visible feedback about this. Something along the lines of this: https://davidwalsh.name/git-hook-npm-install-package-json-modified

Good call on this for sure...

**Comment by gashcrumb** [comment]
On the post-merge hook, created https://github.com/syndesisio/syndesis-ui/issues/620 as it'd be good to make it the default for anyone making a fresh fork of the repo.

**Comment by riccardo-forina** [comment]
Ok, so, tslint hangs forever trying to lint `syndesis-form-control.component.ts`. Trying to figure out why, but looks like there is no verbose output from the tool...

Any chance any of you got a problem like this in the past?

**Comment by riccardo-forina** [comment]
Out of curiosity I left the linter running, with a time on it. Eventually my machine crawled to almost an halt, I managed to snap this:

```
➜  syndesis-ui git:(issue-585) ✗ time yarn lint
yarn lint v0.27.5
$ yarn ng lint
Done in 6750.62s.
yarn lint  5191.37s user 989.64s system 91% cpu 1:52:30.82 total
```

:weary:

gashcrumb:
Wow!  That's a new one, I'll try it here in a few when I'm in front of my
workstation...

riccardo-forina:
Got what the problem was. For laziness I styled the tooltip using some inline styling in the component itself, instead of using a new scss file just for that.
That made the linter go nuts.

Moved the style in its own scss file, all is fine.

So, the moral is *do not use inline styles in directives*.

gashcrumb:
> So, the moral is do not use inline styles in directives.

Good to know!  Seems like a bad bug to have though, I guess we'll have to look and see if there's an open issue.  Anyway, thanks for updating this!  I'll merge this a bit later today as we've a demo to give internally this morning.

PR #627, opened 2017-07-20
fix(step-configure): Fix the data mapper view

### LINKED ISSUE #621: Datamapper Step does not show up
When adding a datamapper step, then nothing changes on the main screen (still shows "Add to integration").

In the console it can be seen that the mapping for `io.syndesis.connector.salesforce.Contact` cannot be found, which is possibly due to that the classpath request returns an empty classpath (as already mentioned in https://github.com/atlasmap/atlasmap/issues/28

However I wonder why atlasmap returns a 200 as HTTP return code and not a 404 ?

gashcrumb:
Also run prettier on step-configure.component.ts

fixes #621

### DIFF
--- a/src/app/integrations/edit-page/step-configure/step-configure.component.ts
+++ b/src/app/integrations/edit-page/step-configure/step-configure.component.ts
@@ -1,4 +1,10 @@
-import { Component, Input, OnInit, OnDestroy, ChangeDetectorRef } from '@angular/core';
+import {
+  Component,
+  Input,
+  OnInit,
+  OnDestroy,
+  ChangeDetectorRef,
+} from '@angular/core';
 import { Subscription } from 'rxjs/Subscription';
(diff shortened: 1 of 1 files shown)


src/app/integrations/edit-page/step-configure/step-configure.component.ts
```diff
@@ -114,28 +120,30 @@ export class IntegrationsStepConfigureComponent extends FlowPage
...
+        log.debugc(() => 'stepConfig: ' + JSON.stringify(stepDef));
         if (!stepDef) {
           // TODO if we don't have a definition for this step then ???
-          switch (step.stepKind) {
-            case 'basic-filter':
-            case 'mapper':
-              log.debugc(() => 'No form configuration, skipping the form building service..');
-              return;
-          }
+          return;
```
kahboom: Good call..

PR #632, opened 2017-07-21
More Filter Stuff

kahboom:
Other filter requirements (and improvements?).


kahboom:
@gashcrumb - Can't remember where I am with this, but would you mind re-reviewing and I'll just create a more specific PR for the new filter-related issues?

PR #656, opened 2017-07-25
fix(auth): Notification toast when refresh token fails

### LINKED ISSUE #217: Add redirect to login page when token times out
Currently only a 401 happens for an Ajax request and the UI doesn't change.

astefanutti:
fixes #217

### DIFF
--- a/src/app/app.component.html
+++ b/src/app/app.component.html
@@ -1,4 +1,7 @@
 <div class="app">
+  <!-- Toast Notification Container -->
+  <toaster-container></toaster-container>
+
   <header *ngIf="loggedIn">
 
     <!-- Navbar: Logged Out -->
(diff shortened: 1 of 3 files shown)


gashcrumb:
@astefanutti have some test failures, most likely you need to add the toaster module to `app.component.spec.ts`.  You can run `yarn test` in a second terminal and they'll run whenever you make a change.

astefanutti:
@gashcrumb thx! Still need to assimilate all that 😃.

jludvice:
btw @jimmidyson  why was I assigned to this issue by @jimmibot  ?

astefanutti:
@jludvice it may be it's based on the commit logs, finding the most active contributor intersecting with the PR changes.

PR #665, opened 2017-07-26
feat(settings): Finish up with modal prompt

### LINKED ISSUE #546: Setup/Settings Page
Right now we have a bit of a mess with where projects get save in github.  It get saved under the last user that published the integration.  If 2 different users save the integration, then it will be save in two different git repos.  I think all integrations should get saved under the same github organization regardless of the user. 

We should have a setup or settings page where users configure:
1. The github organization that integration projects get created under
2. The OpenShift namespace that the project get deployed into

At this point we might want to trigger oauth flows to collect offline tokens which the system then uses when working with github/openshift.

gashcrumb:
Finish up the oauth settings with a modal prompt for removing credentials

fixes #546

screencapture-localhost-4200-settings-oauth-clients-1501078911793

### DIFF
--- a/src/app/settings/oauth-apps/oauth-app-form.component.html
+++ b/src/app/settings/oauth-apps/oauth-app-form.component.html
@@ -3,7 +3,7 @@
     <div class="alert alert-success"
          *ngIf="message">
       <span class="pficon pficon-ok"></span>
-      <strong>Registration Successful!</strong> You can validate these values when <a [routerLink]="['/connections/create']">creating a connection</a> for {{item.client.name}}
+      <strong>Registration Successful!</strong> You can validate these values when <a [routerLink]="['/connections/create']">creating a connection</a>      for {{item.client.name}}
     </div>
     <div class="alert alert-warning"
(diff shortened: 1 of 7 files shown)


src/app/settings/oauth-apps/oauth-app-form.component.html
```diff
@@ -3,7 +3,7 @@
     <div class="alert alert-success"
          *ngIf="message">
       <span class="pficon pficon-ok"></span>
-      <strong>Registration Successful!</strong> You can validate these values when <a [routerLink]="['/connections/create']">creating a connection</a> for {{item.client.name}}
+      <strong>Registration Successful!</strong> You can validate these values when <a [routerLink]="['/connections/create']">creating a connection</a>      for {{item.client.name}}
```
kahboom: What's up with the huge space here?

kahboom:
@gashcrumb - Checks passing before suddenly failed :(

gashcrumb:
Hmm, wonder if this needs to be marked as reviewed again to merge?

Issue #604, opened 2017-07-12
Data Mapper Integration For Multi-Source


jbakermd:
The Data Mapper now supports Multi-Source. This needs to be integrated into Syndesis' UI. 

Example multi-source configuration is shown here:

https://github.com/atlasmap/atlasmap-ui/blob/master/src/app/lib/syndesis-data-mapper/components/data.mapper.example.host.component.ts#L66

gashcrumb:
The data mapper also supported multiple targets all along didn't it?  

I think the main work to be done here is to iterate through the previous steps in the integration and pull out any available `OutputDataShape` attributes, and through any steps after the data mapper step and pull out `InputDataShape` attrbiutes and configure the data mapper with all these.

jbakermd:
The UI technically supports multiple targets. I use that functionality for testing various mapping scenarios fairly often. However, the data mapper engine itself is not scoped to accommodate multiple targets. So, we need to ensure only one target is passed to the data mapper's initial configuration for now.

PR #685, opened 2017-07-27
feat(integrations): Add additional convenience methods

gashcrumb:
relates to #604


src/app/integrations/edit-page/current-flow.service.ts
```diff
@@ -60,7 +60,46 @@ export class CurrentFlow {
     return this.steps.slice(1, -1);
   }
 
-  getPreviousConnections(position): Array<Step> {
+  /**
+   * Return all steps in the flow after the supplied position
```
kahboom: Yay!

PR #686, opened 2017-07-27
feat(data-mapper): Pass multiple source to datamapper if available

### LINKED ISSUE #604: Data Mapper Integration For Multi-Source

gashcrumb:
Also more intelligently grabs previous and subsequent connections out of the integration

fixes #604

### DIFF
--- a/src/app/integrations/edit-page/step-configure/data-mapper/data-mapper-host.component.ts
+++ b/src/app/integrations/edit-page/step-configure/data-mapper/data-mapper-host.component.ts
@@ -21,15 +21,15 @@ import { MappingSerializer } from 'syndesis.data.mapper';
 
 import { DataMapperAppComponent } from 'syndesis.data.mapper';
 
-import { ConfigService } from '../../../config.service';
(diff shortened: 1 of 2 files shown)


src/app/integrations/edit-page/step-configure/data-mapper/data-mapper-host.component.ts
```diff
@@ -21,23 +21,23 @@ import { MappingSerializer } from 'syndesis.data.mapper';
 
 import { DataMapperAppComponent } from 'syndesis.data.mapper';
 
-import { ConfigService } from '../../../config.service';
-import { IntegrationSupportService } from '../../../store/integration-support.service';
+import { ConfigService } from '../../../../config.service';
```
kahboom: We really need to find a better way of dealing with these..

PR #698, opened 2017-07-28
feat(connections): fetch credentials from API when user selects connector

gashcrumb:
Currently just fetches the data and stores it, next step is figure out what to do with it

relates to #556


src/app/connections/create-page/configure-fields/configure-fields.component.html
```diff
@@ -1,3 +1,5 @@
 <div class="wizard-pf-contents">
-  <syndesis-connection-view [(connection)]="connection" [mode]="'edit'" [showName]="false"></syndesis-connection-view>
+  <syndesis-connection-view [(connection)]="connection"
```
kahboom: Ty! :)

Issue #631, opened 2017-07-21
Cannot add additional filter step


rhuss:
@rhuss commented on [Fri Jul 21 2017](https://github.com/syndesisio/syndesis-rest/issues/452)

When pussing "+ Add another rule" the following JavaScript exception occurs. This happens when creating a new filter step or editing an already existing filter step

![image](https://user-images.githubusercontent.com/99080/28450858-49ceacf4-6dea-11e7-8676-c7e33f72215f.png)

kahboom:
@rhuss - Can you still reproduce this?

kahboom:
Update: I'm able to reproduce on staging, but not locally. Looking into this.

rhuss:
I can reproduce it on staging and locally with the latest changes from master.

kahboom:
Can confirm this is an AOT compiling issue with ng2-dynamic-forms, causing a runtime exception. Will try excluding this function causing the error from mangling and see how that works out. Wish me luck...

kahboom:
The `findById` method provided by ng2-dynamic-forms does not appear to work with AOT. I tried the following alternative of directly accessing the object in the array, rather than traversing the DOM with an element selector, which seems to have worked, but does there is some hacking to be done, as it still does not work and renders a different type of error. And this is why I hate working with too many third party JS libraries: 😂

<img width="1031" alt="screenshot 2017-07-28 21 55 53" src="https://user-images.githubusercontent.com/3844502/28741276-c048f0f0-73e0-11e7-855c-9b3312681ded.png">
<img width="1303" alt="screenshot 2017-07-28 21 56 24" src="https://user-images.githubusercontent.com/3844502/28741278-c390c2d8-73e0-11e7-9894-b9ccce1c7dc1.png">
<img width="1309" alt="screenshot 2017-07-28 21 56 38" src="https://user-images.githubusercontent.com/3844502/28741279-c5d29f12-73e0-11e7-87e3-010bc86f5b46.png">

kahboom:
@gashcrumb - This has been a blocker for a while (my fault), but I'm wasting too much time on it and need to move forward. I've been working on other issues simultaneously, but this is slowing me to a crawl and seems very straightforward. If you could take a look, I would really appreciate it.

I have opened up an issue with the ng2-dynamic-forms lib here: https://github.com/udos86/ng2-dynamic-forms/issues/458

Btw, getting the following AOT transpilation errors as of the latest pull (only on AOT):

[...]

PR #710, opened 2017-08-01
WIP re #631 Basic Filter Add Step

kahboom:
### WIP - Do not merge
This is still underway, due to a bug. We might be able to do a workaround with a custom template, but it's a bit hack-y. We can also just keep the Add button in each row/rule, but it wouldn't be consistent with the designs. Perhaps that will be the best for now. Leaving this for last as it is blocking me from working on other tasks, just wanted to have the PR open so @gashcrumb could take a look and let me know of any ideas.

### DIFF
--- a/src/app/app.component.ts
+++ b/src/app/app.component.ts
@@ -41,7 +41,7 @@ export class AppComponent implements OnInit, AfterViewInit {
 
   loggedIn = false;
   title = 'Syndesis';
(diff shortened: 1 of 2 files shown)


gashcrumb:
I think with https://github.com/syndesisio/syndesis-ui/pull/713 we don't need this PR, probably should add the twitter URL change to a separate PR.

PR #739, opened 2017-08-07
feat(connections): Enable filtering by tags in connections list toolbar

astefanutti:


src/app/connections/list-toolbar/list-toolbar.component.ts
```diff
@@ -60,7 +62,21 @@ export class ConnectionsListToolbarComponent<T> implements OnInit, OnDestroy {
     } as ToolbarConfig;
 
     this.subscription = this.items
-      .do(connections => this._allItems = connections)
+      .do(items => this._allItems = items)
+      .do(items => {
+        if (_.some(items, item => _.isArray(item['tags']))) {
```
gashcrumb: You can use ES6 stuff, i.e. `items.filter()`, `items.find()` etc.  I think actually we could get rid of the lodash dependency, I think there's just one bit of code that's using it currently.

src/app/connections/list-toolbar/list-toolbar.component.ts
```diff
@@ -70,11 +86,13 @@ export class ConnectionsListToolbarComponent<T> implements OnInit, OnDestroy {
...
 
   filter(): void {
-     const result = this.toolbarConfig.filterConfig.appliedFilters
-      .reduce((items, filter) => this.propertyFilter.transform(items, {
-          filter       : filter.value,
-          propertyName : filter.field.id,
-        }), this._allItems);
+    const result = this.toolbarConfig.filterConfig.appliedFilters
+      .reduce((items, filter) => filter.field.id === 'tag'
+        ? _.filter(items as Taggable[], { tags: [filter.query.value] })
```
gashcrumb: `items.filter()`

src/app/connections/list-toolbar/list-toolbar.component.ts
```diff
@@ -91,4 +109,12 @@ export class ConnectionsListToolbarComponent<T> implements OnInit, OnDestroy {
     });
     this.filteredItems.next(result);
   }
+
+  filterFieldSelected($event: FilterEvent) {
+    const field = $event.field;
+    if (field.id === 'tag') {
+      field.queries = _.uniq(_.flatMap(this._allItems, item => item['tags'] || []))
```
gashcrumb: hmm, might be a bit trickier without lodash, but:

flatMap => https://stackoverflow.com/questions/39837678/why-no-array-prototype-flatmap-in-javascript
uniq => https://stackoverflow.com/questions/11246758/how-to-get-unique-values-in-an-array

We could probably stand to have some general helper functions for this kind of stuff somewhere, maybe under 'common'

src/app/model.ts
```diff
@@ -6,6 +6,10 @@ export interface BaseEntity {
     kind ? : string;
 }
 
+export interface Taggable {
```
gashcrumb: This file is generated, these changes will likely get lost.  

Generator code is here -> https://github.com/syndesisio/syndesis-ui/tree/master/scripts

another option is to maybe define this interface elsewhere, maybe we need a `model-extensions.ts` file where we can further enrich types as needed, then we'd have places to put  stuff like -> https://github.com/syndesisio/syndesis-ui/blob/master/src/app/store/step/step.store.ts#L4-L14 etc.

src/app/connections/list-toolbar/list-toolbar.component.ts
```diff
@@ -60,7 +62,21 @@ export class ConnectionsListToolbarComponent<T> implements OnInit, OnDestroy {
...
+          if (!_.find(filterConfig.fields, { id: 'tag' })) {
+            filterConfig.fields.push({
+              id          : 'tag',
+              title       : 'Tag',
+              placeholder : 'Filter by tag...',
+              type        : 'typeahead',
+            });
+          }
+        } else {
+          _.remove(filterConfig.fields, { id: 'tag' });
```
gashcrumb: `filterConfigFields.slice()` probably?  Can get the index from `find` if I recall correctly.

astefanutti:
@gashcrumb thanks for the review. Make all sense. I've removed the type thing as it'll make the impl more reusable anyway.

astefanutti:
@gashcrumb it should be updated according to your review by now. Let me know if you have any further points.

PR #753, opened 2017-08-09
refactor(basic-filter): Adapted for custom forms

rhuss:


src/app/store/integration-support.service.ts
```diff
@@ -20,6 +20,7 @@ export class IntegrationSupportService {
   }
 
   getFilterOptions(integration: Integration): Observable<any> {
+    // Should be /{connectorId}/actions/{actionId}/filters/options
```
rhuss: that's a bogus comment

rhuss:
@akieling @gashcrumb could you please have look ?

rhuss:
sorry, please wait a bit. just messed up the PR.

rhuss:
Ok, good to go for a review.

gashcrumb:
@rhuss you've a lint error:

```
src/app/store/step/step.store.ts[198, 27]: expected nospace before colon in call-signature
```

you can run `yarn lint` locally.  Also it's possible to set up your IDE to fix this kind of stuff automatically on save, at least that's how I have vscode set up.  Then I just have to manually make sure I don't commit `console.log` statements all over the code :-)

src/app/store/step/step.store.ts
```diff
@@ -192,4 +192,10 @@ export class StepStore {
   getSteps() {
     return this.steps;
   }
+
+  // Check if we need a custom form handling which stores the parsed
+  // properties in customProperties
+  isCustomStep(step: Step) : boolean {
+    return step.stepKind === BASIC_FILTER;
```
gashcrumb: The data mapper should fall into this category too really.  If we set an empty `configuredProperties` on the step object in that case it shouldn't be the end of the world.  In which case we can reduce the code a bit more in the step-configure controller, WDYT?

rhuss: If we could agree on a custom handling that would be awesome. However I found that the handling was a bit different, e.g. when the datamapper step finishes, a different navigation [happens](https://github.com/rhuss/syndesis-ui/blob/4d0760d8ad5e36845616bff93ed101a17cde1128/src/app/integrations/edit-page/step-configure/step-configure.component.ts#L68)

For https://github.com/rhuss/syndesis-ui/blob/4d0760d8ad5e36845616bff93ed101a17cde1128/src/app/integrations/edit-page/step-configure/step-configure.component.ts#L145 I think we can combine both easily.

PR #756, opened 2017-08-09
feat(integrations): Use a hatchet on configuredProperties values for now

gashcrumb:
relates to syndesisio/syndesis-rest#451

### DIFF
--- a/src/app/integrations/edit-page/step-configure/filter-steps/basic-filter.model.ts
+++ b/src/app/integrations/edit-page/step-configure/filter-steps/basic-filter.model.ts
@@ -108,10 +108,13 @@ export function createBasicFilterModel(configuredProperties: BasicFilter) {
     ];
   };
   let groups = undefined;
+  let rules = undefined;
   // build up the form array from the incoming values (if any)
   if (configuredProperties && configuredProperties.rules && configuredProperties.rules.length) {
+    // TODO hackity hack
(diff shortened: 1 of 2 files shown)


src/app/integrations/edit-page/step-configure/filter-steps/basic-filter.model.ts
```diff
@@ -108,10 +108,13 @@ export function createBasicFilterModel(configuredProperties: BasicFilter) {
     ];
   };
   let groups = undefined;
+  let rules = undefined;
   // build up the form array from the incoming values (if any)
   if (configuredProperties && configuredProperties.rules && configuredProperties.rules.length) {
+    // TODO hackity hack
+    rules = JSON.parse(<any>configuredProperties.rules);
```
rhuss: Not sure whether this is correct in this place, has configuredProperties is of type `BasicFilter` which in turn contains the rules as an array of `Rule` elements already.

gashcrumb:
This bit is for loading the rules *back* from the step, the form expects
this to be an array...

rhuss:
Hmm. But `configuredProperties` is of type `BasicFilter` where rules is an array of `FilterRules`. An actually I indeed got an parse error at this point.

gashcrumb:
`configuredProperties` should be the containing object and `rules` should be the attribute that contains the stringified array.  But lemme check as I couldn't test this yesterday as I couldn't save the filter step still.

PR #773, opened 2017-08-15
feat: Swap to nginx from caddy

### LINKED ISSUE #622: Change HTTP server to nginx instead of caddy
As it can be productised easier. See https://issues.jboss.org/browse/ENTESB-6991 for details.

jimmidyson:
Fixes #622


zregvart:
I wonder if we could use the [openshift-nginx](https://github.com/nginxinc/openshift-nginx) outright (as in Dockerfile `FROM`) and just add our configuration files. Something like:

add to `package.json`:

    docker pull registry.access.redhat.com/rhscl/nginx-110-rhel7

then the `Dockerfile` is:

    FROM registry.access.redhat.com/rhscl/nginx-110-rhel7
    COPY ...

jimmidyson:
AFAIK we can't use RHEL based images for upstream projects...

rhuss:
Technically I would prefer this, too, but share the same concerns as @jimmidyson that, although the image is pullable, its not supposed to be distributed eg. via Docker Hub (which we do, because we push this image to Dockerhub, too, as a binary)

Dockerfile
```diff
@@ -1,5 +1,46 @@
...
+# forward request and error logs to docker log collector
+RUN ln -sf /dev/stdout /var/log/nginx/access.log && \
+    ln -sf /dev/stderr /var/log/nginx/error.log && \
+# Change pid file location & remove nginx user & change port to 8080
+    sed -i 's/\/var\/run\/nginx.pid/\/var\/cache\/nginx\/nginx.pid/g' /etc/nginx/nginx.conf && \
+    sed -i -e '/user/!b' -e '/nginx/!b' -e '/nginx/d' /etc/nginx/nginx.conf && \
+    rm -f /etc/nginx/conf.d/default.conf && \
+# modify perms for non-root runtime
+    chown -R 998 /var/cache/nginx /etc/nginx && \
+    chmod -R g=u /var/cache/nginx /etc/nginx
```
rhuss: cool, didn't know that trick ;-) Unix still surprises me after all these years ....

Dockerfile
```diff
@@ -1,5 +1,46 @@
...
+
+#VOLUME ["/var/cache/nginx"]
+
+EXPOSE 8080 8443
```
rhuss: Doesn't look like that we expose 8443 in the nginx.conf ?

jimmidyson: We don't yet, but now that OpenShift 3.6 properly supports re-encrypted routes to services using generated service serving cerificates, we will do ASAP (when our infra and OSO is at the appropriate release).

rhuss: ok

rhuss:
looks good to me. Is it ok when I move the docker build files into a subdir ? would push directly to this branch before merging.

rhuss:
oops. so sorry !!! just thought you has overlooked my last comment (as you answered the other one before).

wont happen again.

jimmidyson:
@rhuss No worries! Thanks for trying to help!

PR #781, opened 2017-08-16
fix(integration): Add 'cancel' button to 'add step' and 'add connecti…

### LINKED ISSUE #754: No 'cancel' when adding a connection to an integration
When adding a (third) connection to an integration, there is no way to cancel this step:

image

akieling:
…on' pages

Fix #754

### DIFF
--- a/src/app/integrations/edit-page/edit-page.component.spec.ts
+++ b/src/app/integrations/edit-page/edit-page.component.spec.ts
@@ -75,7 +75,7 @@ describe('IntegrationsEditComponent', () => {
     fixture.detectChanges();
   });
 
-  it('should create', () => {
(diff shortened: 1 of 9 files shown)


gashcrumb:
@akieling wonder if rebasing this since I merged @astefanutti's PR would sort the tests...

src/app/integrations/edit-page/edit-page.component.spec.ts
```diff
@@ -75,7 +75,7 @@ describe('IntegrationsEditComponent', () => {
     fixture.detectChanges();
   });
 
-  it('should create', () => {
+  xit('should create', () => {
```
astefanutti: Rebasing on master should work.

PR #815, opened 2017-08-23
Use common modal service for integration flow step deletion

astefanutti:


src/app/common/modal/modal.service.ts
```diff
@@ -19,21 +19,16 @@ export class ModalService {
...
-    return new Promise((resolve, reject) => {
-      const modal = this.registeredModals.get(id);
-      modal.bsModalRef = this.bsModalService.show(modal.template);
-      const subscription = this.bsModalService.onHide.subscribe(event => {
-        subscription.unsubscribe();
-        const result = modal.result;
-        resolve(result);
-      });
-    });
+    const modal = this.registeredModals.get(id);
```
kahboom: So much cleaner! Thank you.

src/app/integrations/edit-page/flow-view/flow-view-step.component.html
```diff
@@ -9,43 +9,10 @@
...
-                  (click)="cancelDeletePrompt()">Cancel</button>
-          <button type="button"
-                  class="btn btn-primary"
-                  (click)="deleteStep()">Delete</button>
-        </div>
-      </div>
-    </div>
-  </div>
+  <syndesis-modal [id]="'delete-integration-step-' + position"
+                  title="Confirm Delete?"
```
kahboom: We could probably define these in the components later on as well, though it's very minor.

PR #863, opened 2017-08-30
fix(integration): Properly deal with connector without action propert…

astefanutti:
…ies definition


kahboom:
Whoops @gashcrumb beat me to it. :)

PR #879, opened 2017-09-01
feat(integrations): send the output data shape instead of the whole i…

### LINKED ISSUE #855: Add a query parameter 'position' when fetching filter options or send outputDataShape
For extracting the possible options for the filter in the BasicFilter, the endpoint `/integrations/filter/options` is called with the current to be defined integration in the POST body.

In order to determine the latest outputShape of the integration, the position of the current filter step within the integration should be sent as a query parameter `position` (where the position is the position is the position the filter step will take).

Alternatively, we could let the UI extract the latest outputShape in the current integration which is part of the latest `action` value before the filter step and only send this outputShape as post payload to the endpoint. Tbh, from a separation-of-concern POV I'd prefer this approach so that the backend only serves what is requested (without deeper logic and might be more easily reused as the datashape is much simpler to send than the whole integration). But I'm fine with either approach.

wdyt, @gashcrumb ?

gashcrumb:
…ntegration

fixes #855 

@rhuss FYI the change doesn't break the view other than the typeahead, but now you can expect to get an object like:
 
```
{
  "kind": "java",
  "type": "twitter4j.Status"
}

```

Instead of the whole integration.

### DIFF
--- a/src/app/integrations/edit-page/step-configure/filter-steps/basic-filter.component.ts
+++ b/src/app/integrations/edit-page/step-configure/filter-steps/basic-filter.component.ts
@@ -98,9 +98,13 @@ export class BasicFilterComponent implements OnChanges {
       self.detector.detectChanges();
     }
 
+    // Get the output data shape from the first previous connection
+    const prevConnection = this.currentFlow.getPreviousConnection(this.position);
+    const dataShape = prevConnection ? prevConnection.action.outputDataShape : {};
+
     // Fetch our form data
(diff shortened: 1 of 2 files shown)


rhuss:
awesome! Unfortunately, I can't update the backend within the next two week. If someone likes to chime in, its in IntegrationHandler.getOptions(), or so. Change should be really easy, just remove the hunt for the outputshape and take it directly from the request. It's only about removing code.

PR #888, opened 2017-09-05
refactor(connection): Remove old connection view page

astefanutti:


gashcrumb:
I think we were going to hold off doing this, but honestly, I don't see us going back to that layout really.  @akieling WDYT?

astefanutti:
@gashcrumb I like removing code 😄!

akieling:
It's fine. I guess we could review the create connection code and see if there is anything from the connection details page that could be reused.

gashcrumb:
@akieling probably worth raising an issue to track that so we don't forget, and I'll go ahead and merge this one...

PR #940, opened 2017-09-12
fix(connection): Fix validation on step 3 of create connection wizard

### LINKED ISSUE #871: Wrong error message on empty name
When trying to save a connection with an empty name I get:

image

The error message should definitely be something like "Please enter a connection name" or so.

### LINKED ISSUE #921: Validate form on step 3 of create connection
**Current behavior**
When the user clicks on the 'Create' button in the final step of the create connection wizard, the form isn't being validated.

**Expected behavior**
Validate the form and show the error messages accordingly.

**Steps to reproduce:**
* Go to step 3 of create connection wizard
* Don't enter a connection name
* Click on the 'Create' button

akieling:
Fixes #921 and fixes #871


gashcrumb:
@akieling compilation failure, wonder if it's a dependency update kinda thing?

Issue #861, opened 2017-08-29
Add JSON schema support for the datamapper source/target document


igarashitm:

igarashitm:
The first commit attached here is just adding an entry point for the JSON schema, nothing else. Let's start here.

datamapper initializes the source/target document here from their DataShape 
https://github.com/syndesisio/syndesis-ui/blob/master/src/app/integrations/edit-page/step-configure/data-mapper/data-mapper-host.component.ts#L184-L201

Right now it doesn't even load nor put the content into createDocumentDefinition(), so only java is allowed for DataShape.type. In order to support JSON schema, we need to load the JSON schema from somewhere, and pass it along.

gashcrumb:
Is this related to https://github.com/syndesisio/syndesis-ui/issues/606 by any chance?

igarashitm:
It looks like so, and it's a datamapper part of the syndesisio/syndesis-project#14. Retrieve a JSON schema from component metadata and pass it into the atlasmap.

igarashitm:
JSON schema inspection in atlasmap has been implemented here - https://github.com/atlasmap/atlasmap/issues/153

igarashitm:
if DataShape.kind == json-instance || json-schema || xml-instance || xml-schema
then we need a content. We'll need a extra key in the DataShape to indicate where the metadata can be retrieved.
https://github.com/igarashitm/syndesis-ui/blob/77e81e1d37689d7fbdb08bff57eae3a41a55f309/src/app/integrations/edit-page/step-configure/data-mapper/data-mapper-host.component.ts#L164-L175

PR #948, opened 2017-09-14
chore: Send JSON schema to mapper

### LINKED ISSUE #947: Is JSON schema provided to the mapper?
We should send `specification` from `DataShape` for `kind` = `json` to the mapper for inspection. Currently we have no properties to map.
screenshot from 2017-09-14 13-26-09

zregvart:
For datashapes with kind=json it is expected that the `specification`
field contains JSON schema as string. This provides the datamapper with
the value therein.

Fixes #947

### DIFF
--- a/src/app/integrations/edit-page/step-configure/data-mapper/data-mapper-host.component.ts
+++ b/src/app/integrations/edit-page/step-configure/data-mapper/data-mapper-host.component.ts
@@ -160,7 +160,7 @@ export class DataMapperHostComponent extends FlowPage
         this.cfg.addJavaDocument(type, isSource);
         break;
       case 'json':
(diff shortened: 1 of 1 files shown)


igarashitm:
Hi @zregvart ,

We now have addJSONSchemaDocument() and addJSONInstanceDocument(), and addJSONDocument() is an alias to addJSONInstanceDocument() to be backward compatible. Not sure if we really need instance document path though, I'd prefer having 'json-schema' and 'json-instance' as kind just like we're doing for XML.

cf.
https://github.com/syndesisio/syndesis-ui/issues/861
https://github.com/igarashitm/syndesis-ui/commit/77e81e1d37689d7fbdb08bff57eae3a41a55f309#diff-2c06191b3f7aa4fb752c7cf65af97673R164

PR #938, opened 2017-09-12
fix(datamapper): src/config.json is ignored for datamapper URL

igarashitm:
Fixes #860

Problem was that the ConfigService instance passed in as a constructor arg is not populated with src/config.json. Once remove this line the shared instance populated with config.json is properly passed in and datamapper url respects them.

PR #951, opened 2017-09-14
fix(datamapper): Add proxy rule to respect datamapper url in the conf…

igarashitm:
…ig.json

Sorry, #938 was not complete to respect datamapper url in the config.json. This proxy rule does it.


igarashitm:
ugh last comma shouldn't be there, yarn start just worked fine though. let me fix it quick

igarashitm:
OK ready to go

PR #969, opened 2017-09-18
fix(navigation): Responsive connection and integration edit pages on …

### LINKED ISSUE #850: Navigation menu is covering up integration workflow
I was in the create an integration workflow and wanted to open up the navigation menu to check on something. 

As the menu opened up, it also covered part of integration workflow interface. My understanding is that opening up the navigation menu would push the integration visualization pane to the right without laying on top of the visualization pane. 

Is this intentional or it's just a bug? 

image

astefanutti:
…nav menu collapse / expand

Fixes #850

### DIFF
--- a/src/app/common/navigation.service.ts
+++ b/src/app/common/navigation.service.ts
@@ -9,17 +9,19 @@ export class NavigationService {
     $.fn.setupVerticalNavigation ? $.fn.setupVerticalNavigation() : '';
   }
 
-  show() {
-    $.fn.setupVerticalNavigation
-      ? $.fn.setupVerticalNavigation().showMenu()
-      : '';
+  expand() {
+    const navElement = $('.nav-pf-vertical');
+    const bodyContentElement = $('.container-pf-nav-pf-vertical');
+
+    navElement.removeClass('collapsed');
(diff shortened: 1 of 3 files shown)


src/app/common/navigation.service.ts
```diff
@@ -9,17 +9,19 @@ export class NavigationService {
...
-  show() {
-    $.fn.setupVerticalNavigation
-      ? $.fn.setupVerticalNavigation().showMenu()
-      : '';
+  expand() {
+    const navElement = $('.nav-pf-vertical');
+    const bodyContentElement = $('.container-pf-nav-pf-vertical');
+
+    navElement.removeClass('collapsed');
+    bodyContentElement.removeClass('collapsed-nav');
```
gashcrumb: Just to double check, does the nav menu behave when you resize the browser in this case?  I thought I had tried this originally but the nav menu maintains some open/close state internally too which necessitated adding the `hideMenu()`/`showMenu()` functions.

astefanutti: Ah let me check!

astefanutti: Damn, you're right. Let me delete that PR and work something out...

PR #992, opened 2017-09-20
fix(integrations): ensure non-typed documents get passed along to the…

### LINKED ISSUE #990: Cannot read property 'initCfg' of undefined
So here I just created a quick oracle EBS to twitter integration and then added a data mapper step.  The data mapper fails to load with the below error:

capture

@akieling @igarashitm FYI...

gashcrumb:
… data mapper

fixes #990

### DIFF
--- a/src/app/integrations/edit-page/step-configure/data-mapper/data-mapper-host.component.ts
+++ b/src/app/integrations/edit-page/step-configure/data-mapper/data-mapper-host.component.ts
@@ -129,28 +129,13 @@ export class DataMapperHostComponent extends FlowPage
   }
 
   createDocumentDefinition(dataShape: DataShape, isSource: boolean = false) {
-    if (!dataShape || !dataShape.type || !dataShape.kind) {
+    if (!dataShape || !dataShape.kind) {
       // skip
       return;
     }
     const type = dataShape.type;
     const kind = dataShape.kind;
-    // avoid passing duplicates off to the data mapper
-    if (isSource) {
-      if (this.sourceDocTypes.find(d => d === type)) {
-        return;
-      }
-      this.sourceDocTypes.push(type);
-    } else {
-      if (this.targetDocTypes.find(d => d === type)) {
-        return;
-      }
-      this.targetDocTypes.push(type);
-    }
-    log.infoc(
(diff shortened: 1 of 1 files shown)


src/app/integrations/edit-page/step-configure/data-mapper/data-mapper-host.component.ts
```diff
@@ -129,45 +129,30 @@ export class DataMapperHostComponent extends FlowPage
...
 
   createDocumentDefinition(dataShape: DataShape, isSource: boolean = false) {
-    if (!dataShape || !dataShape.type || !dataShape.kind) {
+    if (!dataShape || !dataShape.kind) {
       // skip
       return;
     }
     const type = dataShape.type;
     const kind = dataShape.kind;
-    // avoid passing duplicates off to the data mapper
```
igarashitm: I don't know the background though, this duplicate check is no longer required?

src/app/integrations/edit-page/step-configure/data-mapper/data-mapper-host.component.ts
```diff
@@ -129,45 +129,30 @@ export class DataMapperHostComponent extends FlowPage
...
-        return;
-      }
-      this.sourceDocTypes.push(type);
-    } else {
-      if (this.targetDocTypes.find(d => d === type)) {
-        return;
-      }
-      this.targetDocTypes.push(type);
-    }
-    log.infoc(
```
igarashitm: I think we want this log at least?

src/app/integrations/edit-page/step-configure/data-mapper/data-mapper-host.component.ts
gashcrumb: I don't think so, previously there were issues where the component was getting initialized more than once, those should be sorted.

src/app/integrations/edit-page/step-configure/data-mapper/data-mapper-host.component.ts
gashcrumb: I kinda find the logs noisy enough as it is :-)

igarashitm: right? if you say so :)

PR #1022, opened 2017-09-26
fix(assets): rename icon

### LINKED ISSUE #1021: Connections: stored procedure icon references wrong path
Icon file is called `sql-stored.connection.png` in [repository](https://github.com/syndesisio/syndesis-ui/blob/master/src/assets/icons/sql-stored.connection.png).
 
```html
<img _ngcontent-c5="" src="../../../assets/icons/sql-stored-connector.connection.png">
```

gashcrumb:
fixes #1021


syndesisqe:
@gashcrumb just wondering if we need to rename integration icon also.

dsimansk:
@gashcrumb just wondering if we need to rename integration icon also.

gashcrumb:
@dsimansk yup, you're right, pushed...

PR #1030, opened 2017-09-27
fix(integrations): hide tags from integrations until fleshed out

### LINKED ISSUE #1029: Hide Tag related UI elements.
I think we should hide the tags related UI elements right now consider we don't ask users to put in tags when creating integrations. 

image

image

gashcrumb:
fixes #1029


gashcrumb:
circleci seems to be sick this afternoon, needs a retry

PR #1034, opened 2017-09-28
fix: invoke metadata endpoint on action configuration finish

zregvart:
This adds invocation of metadata endpoint when user is done defining
action properties in order to make sure that the input/output data
shapes are set with all the properties defined.

No data shape would be set for action whose properties are defined in
single step otherwise.

Needs error handling.

### DIFF
--- a/src/app/integrations/edit-page/action-configure/action-configure.component.ts
+++ b/src/app/integrations/edit-page/action-configure/action-configure.component.ts
@@ -92,6 +92,24 @@ export class IntegrationsConfigureActionComponent extends FlowPage
       onSave: () => {
         if (!this.lastPage || this.page >= this.lastPage) {
           // all done...
+          this.integrationSupport.fetchMetadata(
+            this.step.connection,
+            this.step.action,
+            this.step.configuredProperties || {},
+          )
+          .toPromise()
+          .then(response => {
+            log.debug('Response: ' + JSON.stringify(response, undefined, 2));
+            const definition: any = response['_body']
+              ? JSON.parse(response['_body'])
+              : undefined;
+            this.step.action.inputDataShape = definition.inputDataShape;
+            this.step.action.outputDataShape = definition.outputDataShape;
+          })
+          .catch(response => {
+            // TODO error handling
+          });
+
           this.router.navigate(['save-or-add-step'], {
             queryParams: { validate: true },
             relativeTo: this.route.parent,


src/app/integrations/edit-page/action-configure/action-configure.component.ts
```diff
@@ -92,6 +92,24 @@ export class IntegrationsConfigureActionComponent extends FlowPage
...
+          .then(response => {
+            log.debug('Response: ' + JSON.stringify(response, undefined, 2));
+            const definition: any = response['_body']
+              ? JSON.parse(response['_body'])
+              : undefined;
+            this.step.action.inputDataShape = definition.inputDataShape;
+            this.step.action.outputDataShape = definition.outputDataShape;
+          })
+          .catch(response => {
+            // TODO error handling
```
gashcrumb: So for this bit if you set this attribute in the class:

https://github.com/zregvart/syndesis-ui/blob/8cb007d9352a7b0fef9fd4b17e69d294a55ffdb3/src/app/integrations/edit-page/action-configure/action-configure.component.ts#L37

like so:

https://github.com/zregvart/syndesis-ui/blob/8cb007d9352a7b0fef9fd4b17e69d294a55ffdb3/src/app/integrations/edit-page/action-configure/action-configure.component.ts#L172-L179

that'll add the inline notification that's in the page already here ->

https://github.com/zregvart/syndesis-ui/blob/8cb007d9352a7b0fef9fd4b17e69d294a55ffdb3/src/app/integrations/edit-page/action-configure/action-configure.component.html#L66-L68

kahboom: Ah cool had no idea we were using inline notifications this way, was just going to use the PatternFly class and an check for showing/hiding tbh lol.

gashcrumb:
@zregvart is this needed for tp1?

zregvart:
> @zregvart is this needed for tp1?

Sorry for late reply, no I don't think so, this only comes up when the last step determines input/output shapes. I don't think we have a connector action that behaves like that for TP1. In TP2 we have Salesforce streaming actions (On Create, On Update) that only have the single step.

Now that I think of it, we should invoke the metadata endpoint before setting up the mapper. This would help with having proper shapes each time the mapper UI is presented.

gashcrumb:
Might be worth raising a separate issue to track that?  Or should we add that to this PR?

I'd probably chain that call with this existing one -> https://github.com/syndesisio/syndesis-ui/blob/master/src/app/integrations/edit-page/step-configure/data-mapper/data-mapper-host.component.ts#L285

zregvart:
> Might be worth raising a separate issue to track that? Or should we add that to this PR?

In the light of that does this PR make sense? Input/output shapes are there only for the mapper, so adding the call before the mapper and after action properties are defined would make the latter superfluous IMHO.

gashcrumb:
Ah, got it, yeah, it totally makes sense to decouple fetching the data shapes from the action configuration really, in fact, we could fetch them from the parent step configuration controller so that it's available to the mapper, basic filter and other steps should they be defined.  So yeah, if we like that approach better then I guess this PR doesn't need to be merged?

zregvart:
I'm happy to close this and work on a better PR as outlined above if someone doesn't beat me to it :racehorse:

gashcrumb:
Cool!  My suggestion would be to start on this page -> https://github.com/syndesisio/syndesis-ui/blob/master/src/app/integrations/edit-page/step-configure/step-configure.component.ts#L187 that's the entry point for building the form and initializing the view, from here I'd make the remote call and defer calling `loadForm()` there until after the data shapes have been fetched.

PR #1036, opened 2017-09-28
WIP: GitHub Auth for FTUE

kahboom:
WIP - Please do not merge. ~Also need to resolve failing tests.~

UXD: https://github.com/syndesisio/syndesis-ux/blob/master/designs/github-settings/github-settings.md

Outstanding:
- ~Collapse sections on completion~.
- ~Update methods to reflect steps more accurately.~
- ~Add static text from designs.~
- ~Step one handling.~
- ~Step two handling.~
  - ~Check if credentials provided are valid.~ (out of scope for TP1)
  - ~HTTP request to GET and POST data from/to Syndesis REST API.~
- Step three handling.
  - ~Re-start login flow.~ Not sure that this is all we need. I need someone to test this out properly.
- ~CSS to reflect completed versus incomplete states.~
- ~CSS to more closely reflect designs (white background, adjust alert).~
- ~Ability to reset setup.~
- ~Error handling and validation in the UI, disable buttons. Success messages.~
- ~Collapse sidebar~.
- ~Update functions for click-handling once we get data from the API.~
- ~Include callback URL~ with copy and paste.

Nice to have:
- Some unit tests, possibly E2E tests or work with QE.
- ~Re-add top navigation bar.~

Screenshots:

screenshot 2017-09-28 10 14 32

cc @akieling @jimmidyson @gashcrumb

### DIFF
--- a/src/app/app.component.html
+++ b/src/app/app.component.html
@@ -1,225 +1,223 @@
-<div class="app" [ngSwitch]="setupService.firstTime">
+<div class="app" [ngSwitch]="userService.checkToken()">
 
-  <!-- Requires GitHub OAuth Setup -->
+    <!-- Does Not Have Valid Token -->
 
-  <ng-container *ngSwitchDefault>
-    <main class="container-fluid setup-required"
-          *ngIf="!loggedIn">
-      <router-outlet></router-outlet>
-    </main>
-  </ng-container>
+    <ng-container *ngSwitchDefault>
+      <main class="container-fluid"
+            *ngIf="!loggedIn">
+        <router-outlet></router-outlet>
+      </main>
+    </ng-container>
 
-  <!-- Does NOT Require GitHub OAuth Setup -->
+    <!-- Has Valid Token -->
 
-  <ng-container *ngSwitchCase="true">
+    <ng-container *ngSwitchCase="true">
 
-    <pfng-toast-notification-list [notifications]="notifications"
-                                  (onActionSelect)="handleAction($event)"
-                                  (onCloseSelect)="handleClose($event)"
-                                  [showClose]="showClose">
-    </pfng-toast-notification-list>
+      <pfng-toast-notification-list [notifications]="notifications"
+                                    (onActionSelect)="handleAction($event)"
+                                    (onCloseSelect)="handleClose($event)"
+                                    [showClose]="showClose">
+      </pfng-toast-notification-list>
 
-    <ng-template #brandTemplate>
-      <a [routerLink]=" ['/'] "
-         class="navbar-brand">
-        <img *ngIf="iconDarkBg && iconDarkBg !== ''"
-             class="navbar-brand-icon"
-             [src]="loggedIn ? iconDarkBg : iconWhiteBg"
-             alt="" />
-        <img *ngIf="logoDarkBg"
-             class="navbar-brand-name"
-             [src]="loggedIn ? logoDarkBg : logoWhiteBg"
-             alt="{{title}}" />
-        <span *ngIf="logoDarkBg === '' && logoWhiteBg === ''"
-              class="navbar-brand-name">{{title}}</span>
-      </a>
-    </ng-template>
-
-    <header>
-
-
-      <!-- Navbar: Logged Out -->
-
-      <nav class="navbar navbar-fixed-top navbar-light loggedOut"
-           *ngIf="!loggedIn">
-        <!-- Logo -->
-        <ng-container *ngTemplateOutlet="brandTemplate"></ng-container>
-      </nav>
-
-
-      <!-- Navbar: Logged In -->
-
-      <nav *ngIf="loggedIn"
-           class="navbar navbar-pf-vertical">
-        <div class="navbar-header">
-
-
-          <!-- Toggle Sidebar-->
-
-          <button type="button"
-                  class="navbar-toggle">
-            <span class="sr-only">Toggle navigation</span>
-            <span class="icon-bar"></span>
-            <span class="icon-bar"></span>
-            <span class="icon-bar"></span>
-          </button>
-
-
-          <!-- Logo -->
-
-          <ng-container *ngTemplateOutlet="brandTemplate"></ng-container>
-        </div>
-
-
-        <!-- Top Navbar -->
-
-        <nav>
-          <ul class="nav navbar-nav navbar-right navbar-iconic">
-
-            <!-- Help Navbar -->
-
-            <li dropdown
-                class="dropdown help">
-              <a dropdownToggle
-                 class="dropdown-toggle nav-item-iconic"
-                 id="dropdownHelp"
-                 aria-haspopup="true"
-                 aria-expanded="true">
-              <span title="Help"
-                    class="fa pficon-help"></span>
-                <span class="caret"></span>
-              </a>
-              <ul class="dropdown-menu"
-                  aria-labelledby="horizontalDropdownMenu1">
-                <li><a href="https://access.redhat.com/documentation/en-us/red_hat_jboss_fuse/7.0-tp/html-single/fuse_ignite_sample_integration_tutorial/"
-
-                       rel="nofollow"
-                       target="_blank">Documentation</a></li>
-                <li><a href="mailto:fuse-ignite-tech-preview@redhat.com"
-                       rel="nofollow"
-                       target="_blank">Contact Us</a></li>
-                <!--
-                <li><a>About</a></li>
-                -->
-              </ul>
-            </li>
-
-            <!-- User w/ No Dropdown -->
-
-            <li class="nav-item-iconic">
-            <span title="Username"
-                  class="fa pficon-user"></span>
-              <span class="username">{{( user | async)?.name}}</span>
-            </li>
-
-            <!-- User Dropdown -->
-
-            <!--
-
-            <li dropdown
-                class="dropdown user">
-              <a dropdownToggle
-                 class="dropdown-toggle nav-item-iconic"
-                 id="dropdownMenu2"
-                 aria-haspopup="true"
-                 aria-expanded="true">
-                <span title="Username"
-                      class="fa pficon-user"></span>
-                <span class="username">{{( user | async)?.name}}</span>
-                <span class="caret"></span>
-              </a>
-              <ul *dropdownMenu
-                  class="dropdown-menu"
-                  aria-labelledby="dropdownMenu2">
-                <li>
-                  <a (click)="resetDB()">Reset DB</a>
-                </li>
-                <li>
-                  <a (click)="exportDB()">Export DB</a>
-                </li>
-                <li>
-                  <a (click)="showImportDB()">Import DB</a>
-                </li>
-              </ul>
-            </li>
-          -->
-          </ul>
-        </nav>
-      </nav>
-      <!--/.navbar-->
-    </header>
-
-
-    <!-- Sidebar -->
-
-    <div class="nav-pf-vertical hidden-icons-pf"
-         *ngIf="loggedIn">
-      <ul class="list-group">
-        <li class="list-group-item"
-            routerLinkActive="active">
-          <a [routerLink]=" ['dashboard'] ">
-          <span class="fa fa-tachometer"
-                data-toggle="tooltip"
-                title="Dashboard"></span>
-            <span class="list-group-item-value">Home</span>
-          </a>
-        </li>
-        <li class="list-group-item"
-            routerLinkActive="active">
-          <a [routerLink]=" ['integrations'] ">
-          <span class="pficon pficon-topology"
-                data-toggle="tooltip"
-                title="Integrations"></span>
-            <span class="list-group-item-value">Integrations</span>
-          </a>
-        </li>
-        <!--
-        <li class="list-group-item" routerLinkActive="active">
-          <a [routerLink]=" ['templates'] ">
-            <span class="pficon pficon-replicator" data-toggle="tooltip" title="Templates"></span>
-            <span class="list-group-item-value">Templates</span>
-          </a>
-        </li>
-        -->
-        <li class="list-group-item"
-            routerLinkActive="active">
-          <a [routerLink]=" ['connections'] ">
-          <span class="pficon pficon-network"
-                data-toggle="tooltip"
-                title="Connections"></span>
-            <span class="list-group-item-value">Connections</span>
-          </a>
-        </li>
-        <li class="list-group-item"
-            routerLinkActive="active">
-          <a [routerLink]=" ['settings'] ">
-          <span class="fa fa-lock"
-                data-toggle="tooltip"
-                title="Settings"></span>
-            <span class="list-group-item-value">Settings</span>
-          </a>
-        </li>
-      </ul>
-    </div>
-
-    <main class="container-fluid container-pf-nav-pf-vertical setup-not-required"
-          *ngIf="loggedIn">
-      <router-outlet></router-outlet>
-    </main>
-
-    <syndesis-modal id="importDb"
-                    title="Import Selection"
-                    [body]="importDbModal">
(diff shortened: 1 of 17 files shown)


kahboom:
Spoke with @dongniwang and @rhuss -- step 3 will no longer be a dropdown, rather just display the user account that has been connected.

src/app/setup/github-oauth.component.html
```diff
@@ -13,15 +17,16 @@
           <div>
             <h2>Register Syndesis</h2>
             <div>
-              <div *ngIf="!stepOneComplete">
+              <div *ngIf="stepOneComplete === false">
```
gashcrumb: Would an `*ngSwitch` and a single variable maybe be better for determining what the current step is?

kahboom: Honestly I think we can refactor with a single var later on, but yes that's likely a better approach.

kahboom:
Should add logout functionality to test, without redirect but just clearing the token and providing it in the top right-hand dropdown, maybe with a toast notification?

kahboom:
@akieling Should we get this merged in and open up a new PR for any remaining changes?

src/app/app.component.html
```diff
@@ -1,13 +1,27 @@
...
-  <ng-template #brandTemplate>
-    <a [routerLink]=" ['/'] "
-       class="navbar-brand">
+<div class="app" [ngSwitch]="userService.checkToken()">
+
+    <!-- Does Not Have Valid Token -->
+
+    <ng-container *ngSwitchDefault>
+      <main class="container-fluid"
+            *ngIf="!loggedIn">
```
gashcrumb: is checkToken() and `loggedIn` the same thing?  Maybe this *ngIf is redundant?

src/app/app.component.html
```diff
@@ -17,178 +31,193 @@
...
 
-          <li class="nav-item-iconic">
-            <span title="Username"
-                  class="fa pficon-user"></span>
-            <span class="username">{{( user | async)?.name}}</span>
-          </li>
+        <!-- Navbar: Logged In -->
 
-          <!-- User Dropdown -->
+        <nav *ngIf="loggedIn"
```
gashcrumb: Same here, redundant or no?

src/app/app.component.html
```diff
@@ -17,178 +31,193 @@
...
-                data-toggle="tooltip"
-                title="Settings"></span>
-          <span class="list-group-item-value">Settings</span>
-        </a>
-      </li>
-    </ul>
+      </div>
+      <main class="container-fluid"
+            [ngClass]="{'container-pf-nav-pf-vertical': !firstTime}"
+            *ngIf="loggedIn">
```
gashcrumb: Same here, is this needed if we're checking the token in a parent element?  I feel like probably not :-)  Might be worth checking if loggedIn could be refactored out altogether.

src/app/app.component.ts
```diff
@@ -135,6 +133,13 @@ export class AppComponent implements OnInit, AfterViewInit {
     this.user = this.userService.user;
     this.notifications = this.notificationService.getNotifications();
     this.showClose = true;
+
+    const apiEndpoint = this.config.getSettings().apiEndpoint;
+    const accessToken = this.oauthService.getAccessToken();
```
gashcrumb: Wonder, do we need a `loading` attribute on the controller and then use the `syndesis-loading` component in the view so that the setup page doesn't flash up until the `isSetupPending()` call is fired?

src/app/app.component.html
akieling: you are right, they are redundant but that's not new code added by this feature

src/app/app.module.ts
```diff
@@ -91,68 +94,93 @@ export function appInitializer(
...
-          }
+          /**
+           * Redirects user to setup page if GitHub setup is still pending
+           */
+          const apiEndpoint = configService.getSettings().apiEndpoint;
+          const accessToken = oauthService.getAccessToken();
+          setupService.isSetupPending(apiEndpoint, accessToken)
+            .then(function (setupPending) {
+              if (setupPending) {
+                window.location.assign('/setup');
```
gashcrumb: This feels like it could be a bit brittle, would it be better to set an attribute on setupService and then do this from the app component's `ngOnInit` function using `router.navigate()` instead?

src/app/app.module.ts
```diff
@@ -91,68 +94,93 @@ export function appInitializer(
...
+                if (autoLinkGithHub && oauthService.state !== 'autolink') {
+                  // Client suggested IDP works great with Keycloak.
+                  oauthService.loginUrl += '?kc_idp_hint=github';
+                  // Clear session storage before trying again.
+                  oauthService.logOut(true);
+                  // And kick off the login flow again.
+                  return oauthService.initImplicitFlow('autolink');
+                }
+
+                init2(oauthService, userService, ngZone, notificationService);
```
gashcrumb: So for this kind of thing where `init2` is only going to be called from this area I'm a bigger fan of just using an inner function, then the relationship between this code and the function being called is a little easier to follow.

src/app/common/user.service.ts
```diff
@@ -2,15 +2,26 @@ import { Injectable } from '@angular/core';
...
+
   get user() {
     return this._user.asObservable();
   }
 
+  /**
+   * Checks to see if user has a token or not
+   */
+  checkToken() {
+    console.log('user.service.ts this.oauthService.hasValidAccessToken(): ' + JSON.stringify(this.oauthService.hasValidAccessToken()));
```
gashcrumb: Honestly should really just change these `console.log` statements to `log.debug` statements, otherwise we start getting lots of console noise.

src/app/integrations/list/list.component.scss
```diff
@@ -1,4 +1,4 @@
-@import 'patternfly-sass/assets/stylesheets/patternfly/color-variables';
+@import '../../../../node_modules/patternfly-sass/assets/stylesheets/patternfly/color-variables';
```
gashcrumb: If we bring in this import we should see what color variables could be used in the file.

src/app/model.ts
```diff
@@ -304,6 +304,15 @@ export interface OAuthApp extends BaseEntity {
...
 
+export interface Setup {
+  gitHubOAuthConfiguration: GitHubOAuthConfiguration;
+}
+
+export interface GitHubOAuthConfiguration {
+  clientId: string;
+  clientSecret: string;
+}
+
```
gashcrumb: If we keep adding to this file we may as well delete the generator code, as this is a generated file.

gashcrumb: Maybe we should create a `client-models.ts` file or something for this kind of thing...

src/app/setup/github-oauth.component.html
```diff
@@ -90,40 +104,55 @@
...
+    <!-- Step Three -->
+
+    <div class="step-three"
+         [ngClass]="{'step-complete' : stepThreeComplete === true}">
+      <div class="col-xs-10">
+        <div class="col-xs-1">
+          <div class="circle pull-right"><p>3</p></div>
+        </div>
+        <div class="col-xs-11">
+          <h2>Configure Integration Location</h2>
```
gashcrumb: I guess this isn't `configure` any more but rather `this is where integrations will be stored` :-)

src/app/setup/github-oauth.component.scss
```diff
@@ -1,6 +1,7 @@
-@import 'patternfly-sass/assets/stylesheets/patternfly/color-variables';
+@import '../../../node_modules/patternfly-sass/assets/stylesheets/patternfly/color-variables';
 
 .github-oauth {
+  background-color: #FFFFFF;
```
gashcrumb: Isn't there a patternfly color variable for this?

src/app/setup/github-oauth.component.scss
```diff
@@ -1,6 +1,7 @@
-@import 'patternfly-sass/assets/stylesheets/patternfly/color-variables';
+@import '../../../node_modules/patternfly-sass/assets/stylesheets/patternfly/color-variables';
```
gashcrumb: Does this need to be referenced as a relative path?  Maybe we can fix the tooling so that's not needed, or maybe we we can have a top-level scss file that imports all of this stuff so that we don't have to have this in every component scss file.

tslint.json
```diff
@@ -35,7 +35,7 @@
     "no-arg": true,
     "no-bitwise": true,
     "no-console": [
-      true,
+      false,
```
gashcrumb: I think we should keep this as-is, it's fine to use `console.log` for debugging purposes, but I'd like to start trimming out logging to the console regularly, and this setting at least ensures you've gotta switch over to logging or take out log statements.

src/app/app.module.ts
```diff
@@ -91,68 +94,93 @@ export function appInitializer(
...
-          ];
-          if (autoLinkGithHub === undefined) {
-            autoLinkGithHub = true;
-          }
+          /**
+           * Redirects user to setup page if GitHub setup is still pending
+           */
+          const apiEndpoint = configService.getSettings().apiEndpoint;
+          const accessToken = oauthService.getAccessToken();
+          setupService.isSetupPending(apiEndpoint, accessToken)
```
gashcrumb: Wonder, could you not do `return setupService.isSetupPending(apiEndpoint, accessToken);` here and then add another promise handler?

That being said, maybe it's best to stick with nesting this call for starters, we can always come back and rework all this initialization code, I feel like as a whole it's gotten a bit unwieldy as we've needed to add functionality.

tslint.json
kahboom: I thought we said it would be okay to remove? Though in retrospect I do think it's probably not good to allow this.

src/app/setup/github-oauth.component.scss
kahboom: This is an absolute pain, I'd love to not have to use relative paths. We can open up an issue for it after TP1.

src/app/setup/github-oauth.component.scss
kahboom: Sure, can change this.

src/app/app.component.html
kahboom: They are the same, but now we are instead making a call to the API to check to see if setup is required and setting the `firstTime` flag, so this `checkToken()` and `ngSwitch` will be going away. This is different because the user is likely to have a token at this point, as they'll be logged in.

src/app/integrations/list/list.component.scss
kahboom: That was the general idea of the `/src/scss/_common.scss` file, or a `main.scss` file, which I think we used to have, but no idea what happened to it. I could add it back in.

src/app/model.ts
kahboom: No opinion on this--it's up to you, just let me know what you prefer.

src/app/setup/github-oauth.component.html
kahboom: I think we're doing away with this step altogether so should be okay! :)

src/app/app.component.ts
kahboom: That would be convenient. I'll see if I can take a look now.

kahboom: This would be easier to do with observables than with promises (which we are currently using). I think it's an enhancement that can be done with TP2. Wdyt?

src/app/app.component.html
kahboom: So, now we are checking for the setup status from the API in addition to the token. I think this app.component.html really needs to be refactored as you say, but perhaps not for TP1.

@akieling Currently the parent element does not take into account the setup status, only whether or not the user has a token, so we are doing that check for individual elements, I guess. There will have to be some redundancy for now.

src/app/app.module.ts
kahboom: I agree. I can try to work on this in the morning, way too late now.

src/app/setup/github-oauth.component.ts
```diff
@@ -59,6 +60,16 @@ export class GitHubOAuthSetupComponent implements OnInit {
   }
 
   /**
+   * Retrieves the callback URL based on the window location HREF.
+   */
+  getCallbackUrl() {
```
kahboom: Of course, here we could do a one-liner or simple `window.location.protocol+"//"+window.location.host + "/"`, but I think keeping it human-readable and less clever is usually better. If there are any objections to that, I'm happy to change it.

kahboom:
screenshot 2017-10-05 06 06 23
screenshot 2017-10-05 06 06 38

src/app/app.component.ts
gashcrumb: Can do it without observables as well, `syndesis-loading` just needs a boolean value.  From a promise you may need to inject `ChangeDetectorRef` and manually trigger change detection.

src/app/model.ts
gashcrumb: Given that the swagger doc is broken anyways at the moment let's not worry about this.  Can you maybe just delete the `generate` build task from the package.json so we can't accidentally overwrite this file?

src/app/setup/github-oauth.component.html
```diff
@@ -1,87 +1,103 @@
-<div class="row github-oauth">
+<div class="row github-oauth"
+     *ngIf="!loading">
+
   <div class="col-xs-12">
     <h1>Getting Started in Syndesis</h1>
```
gashcrumb: Just realized, we can't just use the upstream name on this page, need to inject `ConfigService` and use `{{ config.getSettings('branding', 'appName', 'Syndesis') }}` instead so the correct name can be injected based on the packaging.

For another example see -> https://github.com/syndesisio/syndesis-ui/blob/master/src/app/settings/oauth-apps/oauth-apps.component.html#L4

and inject it like this -> https://github.com/syndesisio/syndesis-ui/blob/master/src/app/settings/oauth-apps/oauth-apps.component.ts#L71

src/app/setup/github-oauth.component.html
```diff
@@ -1,87 +1,103 @@
...
+
     <p>Welcome to Syndesis! To get started you will need to register and authorize Syndesis on GitHub and select a
```
gashcrumb: Same here, we need to inject the app name.

src/app/setup/github-oauth.component.html
```diff
@@ -1,87 +1,103 @@
...
-              </div>
-              <div>
-                <button class="btn btn-primary">Register Syndesis on GitHub</button>&nbsp;<a
-                  href="https://developer.github.com/apps/building-integrations/setting-up-and-registering-oauth-apps/registering-oauth-apps/"
-                  rel="nofollow"
-                  target="_blank">Learn how
-                to register an OAuth app on GitHub.
-                <span class="pficon pficon-info"></span>
+            <div *ngIf="stepOneComplete === false">
+              <p>To start using Syndesis with GitHub, register Syndesis as an OAuth application in GitHub.</p>
```
gashcrumb: And here :-)

src/app/setup/github-oauth.component.html
```diff
@@ -1,87 +1,103 @@
...
+              <p><b>Callback URL: </b> {{callbackUrl}}</p>
+              <button class="btn btn-primary"
+                      (click)="registerSyndesis($event)">Register Syndesis on GitHub
```
gashcrumb: This one too

src/app/setup/github-oauth.component.html
gashcrumb: Yep, might as well delete the commented out stuff really.

akieling:
@gashcrumb I think it's good enough for merging. Please have another look.

src/app/setup/github-oauth.component.scss
akieling: I think webpack can help with that

gashcrumb:
@akieling sounds good, I'll merge and create an issue for the branding fixes.

PR #1046, opened 2017-10-03
fix(assets): Fix icon name (again)

gashcrumb:
Because either @KurtStam or @zregvart apparently can't make up their mind what to call the stored
procedure connector


dsimansk:
If this refers to precreated connection not having an icon displayed. I
believe it's an bug in connectorId in jsondb see
https://github.com/syndesisio/syndesis-rest/issues/650

gashcrumb:
@dsimansk Ah, I see...

@KurtStam @zregvart so then is the connector then called `sql-stored-connector` just to be certain?  Note that changing the `id` on a connector has an immediate impact on the UI as we use that to reference icon file names...

If we're happy with `sql-stored-connector` then this PR can be closed without being merged.

gashcrumb:
OK, connector ID is `sql-stored-connector`

PR #1060, opened 2017-10-06
fix: load data shapes when datamapping step is initialized

### LINKED ISSUE #1055: Cannot edit an existing datamapping
When trying to edit an existing datamapping from a saved integration I get:

image

Target is salesforce.

image

zregvart:
When the datamapper is initialized it needs to be provided with proper
data shapes. Datashapes are a) dynamic and b) sometimes not loaded.
The later (b) is when metadata is fetched for multi-step action on the
next-to-last step (as metadata suggestions are not needed for the last
step), or when configuring a single step action (so metadata suggestion
is fetched without specifying the configured action properties).

This adds a call to fetch metadata and set the input/output data shapes
for the start and end steps, which is limiting and does not handle
errors.

Fetching of the metadata could also be performed in parallel as
data shapes of actions should be independent of each other.

Error handling should be performed.

Fetching metadata should be done for other steps (not just start/end)
and only for the relevant steps (previous/next?).

Fixes #1055

### DIFF
--- a/src/app/integrations/edit-page/step-configure/step-configure.component.ts
+++ b/src/app/integrations/edit-page/step-configure/step-configure.component.ts
@@ -22,6 +22,7 @@ import {
 import { FormFactoryService } from '../../../common/forms.service';
 import { CurrentFlow, FlowEvent } from '../current-flow.service';
 import { Action, Step } from '../../../model';
+import { IntegrationSupportService } from '../../../store/integration-support.service';
 import { log, getCategory } from '../../../logging';
 
 const category = getCategory('IntegrationsCreatePage');
@@ -50,6 +51,7 @@ export class IntegrationsStepConfigureComponent extends FlowPage
     public formService: DynamicFormService,
     public detector: ChangeDetectorRef,
     public stepStore: StepStore,
+    public integrationSupport: IntegrationSupportService,
   ) {
     super(currentFlow, route, router, detector);
   }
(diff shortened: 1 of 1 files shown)


src/app/integrations/edit-page/step-configure/step-configure.component.ts
```diff
@@ -126,6 +128,24 @@ export class IntegrationsStepConfigureComponent extends FlowPage
...
   }
 
+  fetchDataShapesFor(step: Step) {
+    return this.integrationSupport.fetchMetadata(
+      step.connection,
+      step.action,
+      step.configuredProperties || {},
+    )
+    .toPromise()
+    .then(response => {
```
akieling: We try to encapsulate the HTTP call in the service but return the response here. It would be awesome if we just returned the value.

src/app/integrations/edit-page/step-configure/step-configure.component.ts
```diff
@@ -126,6 +128,24 @@ export class IntegrationsStepConfigureComponent extends FlowPage
...
+      log.debug('Response: ' + JSON.stringify(response, undefined, 2));
+      const definition: any = response['_body'] ? JSON.parse(response['_body']) : undefined;
+      step.action.inputDataShape = definition.inputDataShape;
+      step.action.outputDataShape = definition.outputDataShape;
+    })
+    .catch(response => {
```
akieling: Here we could receive a user-friendly error message and use the NotificationService to show it.

gashcrumb:
See https://github.com/syndesisio/syndesis-ui/pull/1061

PR #1068, opened 2017-10-10
fix: set unset action properties to their default values at the end of action definition

zregvart:
Problem: when configuring action properties some of the properties might
depend on previously defined properties, i.e. for Invoke SQL Stored
procedure action `template` parameter depends on `procedureName`
parameter, those properties might end up unset if they are not in in a
step that is one step behind the action property they depend on. This
does not allow for dependencies between properties for actions that have
single step definition.

Solution: set all action property values that are `null` to the
`defaultValue` fetched from metadata endpoint at the point where all
other properties should be defined, i.e. when processing `continue`
handler.


zregvart:
If implementation is not entirely terrible could someone on the UI make it a bit nicer? cc @kahboom @akieling @gashcrumb

PR #1090, opened 2017-10-12
[WIP] feat: Remove keycloak in deference to using an oauth proxy

jimmidyson:
This probably breaks the local development workflow as it requires the oauth proxy, configured with the same cookie secret as the target API server. Thoughts on how to tackle this would be awesome please!

Do not merge until we've sorted the above problem/question.


rhuss:
> This probably breaks the local development workflow as it requires the oauth proxy, configured with the same cookie secret as the target API server. Thoughts on how to tackle this would be awesome please!

The cookie secret is the one generated e.g. on the staging syndesis-rest & atlasmap and is a constant value as long as we do not reapply the templates, right?

What about offering a startup option to provide this secret in a local file, which the app will pickup for the cookie secret if this file exists. That way the secret could be shared and a developer could just write this secret into this file. 

Would this be a possibility ?

jimmidyson:
> The cookie secret is the one generated e.g. on the staging syndesis-rest & atlasmap and is a constant value as long as we do not reapply the templates, right?

Correct.

> What about offering a startup option to provide this secret in a local file, which the app will pickup for the cookie secret if this file exists. That way the secret could be shared and a developer could just write this secret into this file.

Yes that would work and can provide a script to retrieve that cookie secret from target server perhaps.Let me have a play around.

rhuss:
@jimmidyson so good to merge ?

jimmidyson:
Well i still haven't figured out a local development workflow properly... so probably best to hold off?

rhuss:
Ok, I will try it out with the image build from this fork.

I guess the biggest challenge is to get the OpenShift proxy running locally so that the UI can use it ?

chirino:
I've started a s2i image that bakes in all the mvn deps into the ~/.m2/repository directory at:
https://github.com/syndesisio/syndesis-s2i-image
https://hub.docker.com/r/syndesis/syndesis-s2i/

I think this should help shave off some time off the build.  Still need to update the templates to use this instead of the FIS s2i image.

chirino:
@rhuss PR #1105 allows you to do local dev against minishift.  Which was documented in the README as the primary way of getting started with development.

PR #1105, opened 2017-10-16
Updating the developer mode getting started instructions so that it works via redirecting the syndesis-ui service deployed in minishift.

chirino:
This allows you to run in dev mode even when your using an oauth-proxy in front of the UI service.

### DIFF
--- a/package.json
+++ b/package.json
@@ -84,6 +84,7 @@
     "karma-jasmine": "1.1.0",
     "karma-junit-reporter": "1.2.0",
     "karma-mocha-reporter": "2.2.3",
+    "my-local-ip": "^1.0.0",
     "ncp": "^2.0.0",
     "patternfly-sass": "3.23.2",
     "prettier": "^1.5.2",
--- a/package.json
+++ b/package.json
@@ -9,6 +9,7 @@
     "copy:assets": "yarn ncp node_modules/syndesis.data.mapper/src/assets/dm src/assets/dm",
     "start": "yarn ng serve -- --ssl true --ssl-key localhost.key --ssl-cert localhost.crt --proxy-config proxy.conf.js",
     "start:prod": "yarn ng serve -- --ssl true --ssl-key localhost.key --ssl-cert localhost.crt --proxy-config proxy.conf.js --aot --prod",
+    "start:minishift": "yarn ng serve -- --host 0.0.0.0 --disable-host-check",
     "lint": "yarn ng lint",
     "lint:fix": "yarn ng lint -- --fix",
     "test": "yarn ng test",
--- a/scripts/minishift-setup.sh
+++ b/scripts/minishift-setup.sh
@@ -8,7 +8,9 @@ cd "$( dirname "${BASH_SOURCE[0]}" )"
 cd ..
 
 # Lets try to detect the local machine ip..
-LOCAL_IP="$(dig $(hostname) +short)"
+LOCAL_IP="$(./node_modules/.bin/my-local-ip)"
+echo "DETECTED Local IP is: ${LOCAL_IP}"
+
 LOCAL_IP="${1:-$LOCAL_IP}"
 
 sed "s/192.168.64.2/$(minishift ip)/" src/config.json.minishift > src/config.json
--- a/scripts/minishift-setup.sh
+++ b/scripts/minishift-setup.sh
@@ -1,0 +1,66 @@
+
+#!/bin/bash
+#
+#
+
+set -euo pipefail
+cd "$( dirname "${BASH_SOURCE[0]}" )"
+cd ..
+
+# Lets try to detect the local machine ip..
+LOCAL_IP="$(nslookup $(hostname) | tail -n +4 | grep Address | cut -d " " -f 2)"
+LOCAL_IP="${1:-$LOCAL_IP}"
+
+sed "s/192.168.64.2/$(minishift ip)/" src/config.json.minishift > src/config.json
+oc replace --force -f - <<EOF
+{
+  "kind": "List",
+  "apiVersion": "v1",
+  "items": [
+    {
+      "apiVersion": "v1",
+      "kind": "Service",
+      "metadata": {
+        "labels": {
+          "app": "syndesis",
+          "component": "syndesis-ui"
+        },
+        "name": "syndesis-ui"
+      },
+      "spec": {
+        "ports": [
+          {
+            "port": 80,
+            "protocol": "TCP",
+            "targetPort": 4200,
+            "name": "http"
+          }
+        ],
+        "selector": {}
+      }
+    },
+    {
+      "apiVersion": "v1",
+      "kind": "Endpoints",
+      "metadata": {
+        "name": "syndesis-ui"
+      },
+      "subsets": [
+        {
+          "addresses": [
+            {
+              "ip": "${LOCAL_IP}"
+            }
+          ],
+          "ports": [
+            {
+              "name": "http",
+              "port": 4200
+            }
+          ]
+        }
+      ]
+    }
+  ]
+}
+EOF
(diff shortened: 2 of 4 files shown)


scripts/minishift-setup.sh
```diff
@@ -0,0 +1,66 @@
...
+
+set -euo pipefail
+cd "$( dirname "${BASH_SOURCE[0]}" )"
+cd ..
+
+# Lets try to detect the local machine ip..
+LOCAL_IP="$(nslookup $(hostname) | tail -n +4 | grep Address | cut -d " " -f 2)"
+LOCAL_IP="${1:-$LOCAL_IP}"
+
+sed "s/192.168.64.2/$(minishift ip)/" src/config.json.minishift > src/config.json
```
jimmidyson: Probably better off with sed expression like:

`"s/\(\"apiBase\": \"https:\/\/\).*\",$/\1${LOCAL_IP}.nip.io\",/"`

scripts/minishift-setup.sh
```diff
@@ -0,0 +1,66 @@
...
+      "apiVersion": "v1",
+      "kind": "Endpoints",
+      "metadata": {
+        "name": "syndesis-ui"
+      },
+      "subsets": [
+        {
+          "addresses": [
+            {
+              "ip": "${LOCAL_IP}"
```
jimmidyson: I love this idea! Use an external service! Nice thinking :bat:man

dsimansk: And can't be the same behaviour achieved by `NodePort` [service](https://kubernetes.io/docs/concepts/services-networking/service/#type-nodeport)?

jimmidyson: That might also work and get rid of the need to specify the IP address at all - genius!

jimmidyson: Ah... node would be the minishift VM, not the local dev machine.

scripts/minishift-setup.sh
```diff
@@ -0,0 +1,66 @@
...
+#!/bin/bash
+#
+#
+
+set -euo pipefail
+cd "$( dirname "${BASH_SOURCE[0]}" )"
+cd ..
+
+# Lets try to detect the local machine ip..
+LOCAL_IP="$(nslookup $(hostname) | tail -n +4 | grep Address | cut -d " " -f 2)"
```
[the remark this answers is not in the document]
chirino: :(

Do you have something that does work for you?

package.json
```diff
@@ -54,7 +55,7 @@
     "zone.js": "0.8.4"
   },
   "devDependencies": {
-    "@angular/cli": "1.2.8",
+    "@angular/cli": "1.4.7",
```
gashcrumb: This change is probably what's making the CI build fail, is this update needed for the `--disable-host-check` flag?

Should be able to reproduce the error locally via `yarn ng build -- --aot --prod`:

```
ERROR in ./src/polyfills.ts
Module build failed: Error: AotPlugin was detected but it was an instance of the wrong class.
This likely means you have several @ngtools/webpack packages installed. You can check this with `npm ls @ngtools/webpack`, and then remove the extra copies.
    at Object.ngcLoader (/workspace/node_modules/@ngtools/webpack/src/loader.js:358:19)
 @ multi ./src/polyfills.ts
ERROR in ./src/main.ts
Module build failed: Error: AotPlugin was detected but it was an instance of the wrong class.
This likely means you have several @ngtools/webpack packages installed. You can check this with `npm ls @ngtools/webpack`, and then remove the extra copies.
    at Object.ngcLoader (/workspace/node_modules/@ngtools/webpack/src/loader.js:358:19)
 @ multi ./src/main.ts
ERROR in /workspace/node_modules/syndesis.data.mapper/src/app/lib/syndesis-data-mapper/components/document.field.detail.component.ts (240,9): Type 'this' has no properties in common with type 'Component'.
ERROR in /workspace/node_modules/syndesis.data.mapper/src/app/lib/syndesis-data-mapper/components/document.field.detail.component.ts (285,9): Type 'this' has no properties in common with type 'Component'.
ERROR in /workspace/node_modules/syndesis.data.mapper/src/app/lib/syndesis-data-mapper/components/namespace.list.ts (137,9): Type 'this' has no properties in common with type 'Component'.
ERROR in /workspace/node_modules/syndesis.data.mapper/src/app/lib/syndesis-data-mapper/components/document.definition.component.ts (248,9): Type 'this' has no properties in common with type 'Component'.
ERROR in /workspace/node_modules/syndesis.data.mapper/src/app/lib/syndesis-data-mapper/components/mapping/mapping.detail.component.ts (285,9): Type 'this' has no properties in common with type 'Component'.
ERROR in /workspace/node_modules/syndesis.data.mapper/src/app/lib/syndesis-data-mapper/components/mapping/mapping.detail.component.ts (301,9): Type 'this' has no properties in common with type 'Component'.
ERROR in /workspace/node_modules/syndesis.data.mapper/src/app/lib/syndesis-data-mapper/components/mapping/transition.selection.component.ts (109,9): Type 'this' has no properties in common with type 'Component'.
ERROR in /workspace/node_modules/syndesis.data.mapper/src/app/lib/syndesis-data-mapper/components/toolbar.component.ts (164,9): Type 'this' has no properties in common with type 'Component'.
ERROR in /workspace/node_modules/syndesis.data.mapper/src/app/lib/syndesis-data-mapper/components/namespace.list.ts (164,9): Type 'this' has no properties in common with type 'Component'.
```

scripts/minishift-setup.sh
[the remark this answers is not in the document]
rhuss: @chirino @jimmidyson :

```
$(dig $(hostname) +short)
```

?

package.json
gashcrumb: If you want, I can look at opening a separate PR that updates angular-ci again and sort out whatever dependency changes would be needed for it to work.

scripts/minishift-setup.sh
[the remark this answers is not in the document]
gashcrumb: yeah, probably `nslookup` isn't the best tool, maybe the `curl` method in this article -> https://www.if-not-true-then-false.com/2010/linux-get-ip-address/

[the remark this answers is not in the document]
rhuss: But this would yield the IP of my router who does the NAT to the external world. Don't think this works here.

package.json
gashcrumb: yeah, in fact `start:minishift` seems to still work if I revert this line locally, I'd recommend setting this version back to 1.2.8 for this PR, am planning on updating dependencies across the board this sprint and it's non-trivial :-)

chirino: The --disable-host-check is needed to make connecting from minishift into the dev server work as the hostname won't be 127.0.0.1 or localhost.  

Doing a normal test bringing up the browser against localhost:4200 is not sufficient.  You have to use ${machine_ip}:4200 to make sure that --disable-host-check is working.

--disable-host-check did not work for me at the angular cli 1.2.8.

chirino: actually it did not work at 1.0.3 but does at 1.2.8 
updating PR.

scripts/minishift-setup.sh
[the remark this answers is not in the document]
chirino: switching to dig.  Works on my machine too.

scripts/minishift-setup.sh
```diff
@@ -0,0 +1,66 @@
...
+#!/bin/bash
+#
+#
+
+set -euo pipefail
+cd "$( dirname "${BASH_SOURCE[0]}" )"
+cd ..
+
+# Lets try to detect the local machine ip..
+LOCAL_IP="$(dig $(hostname) +short)"
```
gashcrumb: yeah, this doesn't work for me here.  Maybe we should just leave this as an input, i.e. I don't mind setting it in my `.bashrc` for example.  On linux though there's `hostname -I | awk '{print $1;}'` but not sure that works on macs

Issue #712, opened 2017-08-01
Use precomputed java inspection when available.


chirino:
Once the following PRs get merged: 
https://github.com/syndesisio/syndesis-rest/pull/481
https://github.com/syndesisio/syndesis-openshift-templates/pull/65

the syndesis rest api will support get precomputed java inspection at the following API endpoint:
`GET /mapper/v1/java-inspections/{connector_id}/{java_class_name}.json`

404 will be returned if a precomputed inspection is not available. 

The UI should feed these precomputed inspection into the datamapper when available.

chirino:
See: https://github.com/atlasmap/atlasmap-ui/issues/51 for info on how to pass the precomuted inspection into the datamapper.

chirino:
PRs merged and service is available on staging.  Example endpoint:

`https://syndesis-staging.b6ff.rh-idev.openshiftapps.com/mapper/v1/java-inspections/twitter/twitter4j.Status.json`

gashcrumb:
Let's see if we can get this in place by the end of this sprint perhaps?

rhuss:
Actually I really believe that its already available (somehow). The pom.xml which is retrieved was syndesis-rest was feed to atlasmap-ui which then called out to Maven previously. This for sure doesnt' happen anymore, or ? So fetching the pom.xml is useless imo, but I need someone to verify this.

igarashitm:
Just to record I found, the integration POM was supposed to be used for java inspection, for the class which is not in the atlasmap runtime classpath. Right now we have all the syndesis connectors in atlasmap spring-boot runtime dependencies, which should be removed at some point. This is why it's working without maven classpath right now. One of maven classpath or binary upload (required for custom connector) will be eventually needed when we remove connectors from atlasmap runtime classpath. Or rely on precomputed inspection via maven and completely disable on demand java inspection.

PR #1132, opened 2017-10-24
fix(datamapper): Add default mapper URL in case mapperEndpoint is not…

igarashitm:
… defined

Follows up: #712

### DIFF
--- a/src/app/app.module.ts
+++ b/src/app/app.module.ts
@@ -1,4 +1,5 @@
 import { APP_INITIALIZER, NgModule, NgZone } from '@angular/core';
+import { InjectionToken } from '@angular/core';
 import { ReactiveFormsModule } from '@angular/forms';
 import { FormsModule } from '@angular/forms';
 import { HttpModule } from '@angular/http';
@@ -16,7 +17,7 @@ import {
   TypeaheadModule,
 } from 'ngx-bootstrap';
 import { TagInputModule } from 'ngx-chips';
-import { RestangularModule } from 'ngx-restangular';
+import { Restangular, RestangularModule } from 'ngx-restangular';
 import { NotificationModule, NotificationService } from 'patternfly-ng';
 import { DataMapperModule } from '@atlasmap/atlasmap.data.mapper';
 
@@ -57,6 +58,17 @@ export function restangularProviderConfigurer(
       return [];
     }
     return data;
+  });
+}
+
+export const RESTANGULAR_MAPPER = new InjectionToken<Restangular>('restangularMapper');
+export function mapperRestangularProvider(
+  restangular: Restangular,
+  config: ConfigService,
+) {
+  return restangular.withConfig((restangularConfigurer) => {
+    const mapperEndpoint = config.getSettings().mapperEndpoint;
+    restangularConfigurer.setBaseUrl(mapperEndpoint ? mapperEndpoint : '/mapper/v1');
   });
 }
 
@@ -110,6 +122,11 @@ export function restangularProviderConfigurer(
       useFactory: appInitializer,
       deps: [ConfigService],
       multi: true,
+    },
+    {
+      provide: RESTANGULAR_MAPPER,
+      useFactory: mapperRestangularProvider,
+      deps: [Restangular, ConfigService],
     },
     ConfigService,
     UserService,
--- a/src/app/store/integration-support.service.ts
+++ b/src/app/store/integration-support.service.ts
@@ -1,27 +1,32 @@
-import { Injectable } from '@angular/core';
-import { Http } from '@angular/http';
+import { Injector, Injectable } from '@angular/core';
+import { Http, Response } from '@angular/http';
 import { Restangular } from 'ngx-restangular';
 import { Observable } from 'rxjs/Observable';
+import { ErrorObservable } from 'rxjs/observable/ErrorObservable';
 
+import { RESTANGULAR_MAPPER } from '../app.module';
 import { ConfigService } from '../config.service';
 import { Action, Connection, Integration } from '../model';
+import { log, getCategory } from '../logging';
 
 @Injectable()
 export class IntegrationSupportService {
-  mapperBaseUrl: string = undefined;
   service: Restangular = undefined;
   filterService: Restangular = undefined;
   metadataService: Restangular = undefined;
+  mapperService: Restangular = undefined;
   configService: ConfigService = undefined;
 
-  constructor(public restangular: Restangular,
-              public http: Http,
-              public config: ConfigService) {
+  constructor(restangular: Restangular,
+              injector: Injector,
+              private http: Http,
+              private config: ConfigService) {
     this.service = restangular.service('integration-support');
     this.filterService = restangular.service('integrations');
     this.metadataService = restangular.service('connections');
+    const restangularMapper = injector.get(RESTANGULAR_MAPPER);
+    this.mapperService = restangularMapper.service('java-inspections');
     this.configService = config;
-    this.mapperBaseUrl = config.getSettings().apiBase + config.getSettings().mapperEndpoint;
   }
 
   getFilterOptions(dataShape: any): Observable<any> {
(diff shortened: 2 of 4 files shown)


src/app/store/integration-support.service.ts
```diff
@@ -41,8 +47,11 @@ export class IntegrationSupportService {
     return this.http.post(url, configuredProperties);
   }
 
-  requestJavaInspection(connectorId: String, type: String) {
-    const url = this.mapperBaseUrl + '/java-inspections/' + connectorId + '/' + type + '.json';
-    return this.http.get(url);
+  requestJavaInspection(connectorId: String, type: String): Observable<Response> {
+    if (this.mapperBaseUrl) {
+      const url = this.mapperBaseUrl + '/java-inspections/' + connectorId + '/' + type + '.json';
```
gashcrumb: So with this much URL assembling it'd be probably better to use a helper to build the URL, for example:

https://github.com/syndesisio/syndesis-ui/blob/master/src/app/store/integration-support.service.ts#L40

I kinda wonder if we shouldn't consider bringing in https://medialize.github.io/URI.js/

src/app/store/integration-support.service.ts
```diff
@@ -21,7 +23,11 @@ export class IntegrationSupportService {
     this.filterService = restangular.service('integrations');
     this.metadataService = restangular.service('connections');
     this.configService = config;
-    this.mapperBaseUrl = config.getSettings().apiBase + config.getSettings().mapperEndpoint;
+    const apiBase = config.getSettings().apiBase;
```
gashcrumb: I'm not sure `apiBase` is the right setting.  I thought the real setting for the rest API is `apiEndpoint`, at least a quick check on staging yeilds:

```
{
  "apiEndpoint": "https://syndesis-staging.b6ff.rh-idev.openshiftapps.com/api/v1",
  "title": "Syndesis",
  "datamapper": {
    "baseJavaInspectionServiceUrl": "https://syndesis-staging.b6ff.rh-idev.openshiftapps.com/v2/atlas/java/",
    "baseXMLInspectionServiceUrl": "https://syndesis-staging.b6ff.rh-idev.openshiftapps.com/v2/atlas/xml/",
    "baseJSONInspectionServiceUrl": "https://syndesis-staging.b6ff.rh-idev.openshiftapps.com/v2/atlas/json/",
    "baseMappingServiceUrl": "https://syndesis-staging.b6ff.rh-idev.openshiftapps.com/v2/atlas/"
  },
  "oauth": {
    "clientId": "syndesis-ui",
    "scopes": ["openid"],
    "oidc": true,
    "hybrid": true,
    "issuer": "https://syndesis-staging.b6ff.rh-idev.openshiftapps.com/auth/realms/syndesis",
    "auto-link-github": true
  },
  "branding": {
    "logoWhiteBg": "assets/images/syndesis-logo-svg-white.svg",
    "logoDarkBg": "assets/images/syndesis-logo-svg-white.svg",
    "iconWhiteBg": "assets/images/glasses_logo_square.png",
    "iconDarkBg": "assets/images/glasses_logo_square.png",
    "appName": "Syndesis",
    "favicon32": "/favicon-32x32.png",
    "favicon16": "/favicon-16x16.png",
    "touchIcon": "/apple-touch-icon.png"
 }
}
```

The `apiBase` attribute was added when the local dev proxy was added to support going through oauth pages for twitter/salesforce for development instances.  So I'd either suggest we build this URL off of `apiEndpoint` or just add a whole new `mapperEndpoint` setting that contains the full URL of where the client should go to and then we get rid of this other URL parsing :-)

src/app/store/integration-support.service.ts
igarashitm: Thanks for the comment! can you teach me a bit around it, so the Restangular passed in as a constructor arg seems to be pointing
```https://$(syndesis-rest)/api/v1```
But the java inspection URL is
```https://$(syndesis-rest)/mapper/v1```
is it something I can modify on this existing Restangular instance, or need to instantiate another one here? don't know how to neither...

Comment (gashcrumb) [reply src/app/store/integration-support.service.ts]: In fact yeah, I'd just suggest using restangular to build this URL instead, there's examples in the same file for other API calls.

File: src/app/store/integration-support.service.ts (hunk 2 above)
Comment (gashcrumb) [reply src/app/store/integration-support.service.ts]: I mean, my point is the `config.json` that's part of the openshift templates doesn't contain `apiBase` -> https://github.com/syndesisio/syndesis-openshift-templates/blob/master/generator/03-syndesis-ui.yml.mustache#L106

If we add `mappingEndpoint` then it needs to be added to this file too.

File: src/app/store/integration-support.service.ts (hunk 1 above)
Comment (gashcrumb) [reply src/app/store/integration-support.service.ts]: @igarashitm yeah, I'm kinda thinking maybe `yarn add urijs` and then you can do:

```
   // maybe make mapperUrl a member variable
   const mapperUri = new URI(config.getSettings().apiEndpoint).path('/mapper/v1');
```

and then later on:

```
   const targetUrl = mapperUri.segment('java-inspections').segment(connectorId).segment(type + '.json');
```

File: src/app/store/integration-support.service.ts [hunk 3]
```diff
@@ -1,27 +1,31 @@
...
 export class IntegrationSupportService {
-  mapperBaseUrl: string = undefined;
   service: Restangular = undefined;
   filterService: Restangular = undefined;
   metadataService: Restangular = undefined;
+  mapperService: Restangular = undefined;
   configService: ConfigService = undefined;
 
-  constructor(public restangular: Restangular,
+  constructor(@Inject(Restangular) public restangular: Restangular,
```
kahboom: Should be:
```
constructor(@Inject(Restangular) public restangular Restangular,
@Inject(RESTANGULAR_MAPPER) public restangularMapper Restangular,
```

src/app/store/integration-support.service.ts
```diff
@@ -1,27 +1,31 @@
...
+              @Inject(RESTANGULAR_MAPPER) public restangularMapper: Restangular,
```
gashcrumb: Okay, take out this line and instead inject `injector: Injector`.  This will require adding an `Injector` import from `'@angular/core'` you can add that to line 1 in the file:

```
import { Inject, Injectable, Injector } from '@angular/core';
```

Then in the constructor you can call `injector.get(RESTANGULAR_MAPPER)` which will return the configured restangular instance looks like.  So the full change I have here is:

```
constructor(restangular: Restangular,
              injector: Injector,
              private http: Http,
              private config: ConfigService) {
    this.service = restangular.service('integration-support');
    this.filterService = restangular.service('integrations');
    this.metadataService = restangular.service('connections');
    const restangularMapper = <Restangular> injector.get(RESTANGULAR_MAPPER);
    this.mapperService = restangularMapper.service('java-inspections');
    this.configService = config;
  }
```

src/app/app.module.ts
```diff
@@ -60,6 +61,17 @@ export function restangularProviderConfigurer(
   });
 }
 
+export const RESTANGULAR_MAPPER = new InjectionToken<any>('restangularMapper');
```
gashcrumb: Change this to:

```
export const RESTANGULAR_MAPPER = new InjectionToken<Restangular>('MapperRestangular');
```

gashcrumb: Actually the token value can be `restangularMapper` that's fine, just change the type from `any` to `Restangular` as this token is apparently supposed to be typed.

src/app/store/integration-support.service.ts
kahboom: Thanks @gashcrumb - talk about a workaround.

igarashitm:
@gashcrumb thanks! verified it works as expected.
* mapperEndpoint: '/mapper/v1' => successfully fetched precomputed inspection
* mapperEndpoint: '/mapper/v2' => emit a warn that precomputed inspection could not be fetched. It goes then online inspection
* no mapperEndpoint in config.json => successfully fetched precomputed inspection from /mapper/v1

PR #1165, opened 2017-11-01
feat: Add integration Import/Export buttons.

### LINKED ISSUE #1160: Add integration Import/Export buttons
Integration details view should have an export button.
Integration list should have an import button.
They should use the APIs created with https://github.com/syndesisio/syndesis-rest/issues/764 to download a .zip file which then you pick to import.

chirino:
Fixes #1160.

### DIFF
--- a/src/app/integrations/detail-page/detail.component.ts
+++ b/src/app/integrations/detail-page/detail.component.ts
@@ -5,6 +5,8 @@ import {
   OnDestroy,
   ChangeDetectorRef,
 } from '@angular/core';
+
+import { ResponseContentType , Http } from '@angular/http';
 import { ActivatedRoute, Params, Router, UrlSegment } from '@angular/router';
 import { Observable } from 'rxjs/Observable';
 import { Subscription } from 'rxjs/Subscription';
@@ -14,7 +16,9 @@ import { Integration, Step, Connection, Action } from '../../model';
 import { IntegrationStore } from '../../store/integration/integration.store';
 import { IntegrationViewBase } from '../components/integrationViewBase.component';
 import { ModalService } from '../../common/modal/modal.service';
+import { ConfigService } from '../../config.service';
 import { NotificationService, NotificationType } from 'patternfly-ng';
+import { saveAs } from 'file-saver';
 
 @Component({
   selector: 'syndesis-integration-detail-page',
@@ -136,6 +140,8 @@ export class IntegrationsDetailComponent extends IntegrationViewBase
     public notificationService: NotificationService,
     public modalService: ModalService,
     public application: ApplicationRef,
+    private http: Http,
+    private config: ConfigService,
   ) {
     super(store, route, router, notificationService, modalService, application);
     this.integration = this.store.resource;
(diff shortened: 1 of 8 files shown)


src/app/integrations/detail-page/detail.component.ts
```diff
@@ -402,4 +408,12 @@ export class IntegrationsDetailComponent extends IntegrationViewBase
     this.integrationSubscription.unsubscribe();
     this.routeSubscription.unsubscribe();
   }
+
+  exportIntegration() {
```
gashcrumb: I think this should probably go into the integration-support service...

chirino: Yeah. probably should.

src/app/integrations/detail-page/detail.component.ts
```diff
@@ -402,4 +408,12 @@ export class IntegrationsDetailComponent extends IntegrationViewBase
...
+    const id = this.i.id;
+    const url = this.config.getSettings().apiEndpoint + '/integrations/' + id + '/export.zip';
```
gashcrumb: It'd be better to use restangular to build this URL really, for example:

https://github.com/syndesisio/syndesis-ui/blob/master/src/app/store/integration-support.service.ts#L33

as building URLs this way can get broken really easily...

src/app/integrations/list-page/list-page.component.html
```diff
@@ -4,15 +4,72 @@
...
+  <ng-template #importIntegrationModal let-modal="modal">
+
+    Select the integration file to import:
+    <input #fileSelect type="file" ng2FileSelect [uploader]="uploader" (onFileSelected)="fileSelect.value=''" />
+
+    <!-- for more info/examples of how to use the ng2 file uploader, see https://valor-software.com/ng2-file-upload/ -->
+    <div ng2FileDrop
+      [uploader]="uploader"
+      class="well my-drop-zone">
+      Or drag and drop your integration files here
```
gashcrumb: Awesome :-)

src/app/integrations/list-page/list-page.component.ts
```diff
@@ -29,12 +37,54 @@ export class IntegrationsListPage implements OnInit {
...
+    private modalService: ModalService,
+    private integrationSupportService: IntegrationSupportService,
+    public notificationService: NotificationService,
+    public application: ApplicationRef,
+    private http: Http,
+    private config: ConfigService,
+  ) {
     this.integrations = this.store.list;
     this.loading = this.store.loading;
+    const url = this.config.getSettings().apiEndpoint + '/integration-support/import';
```
gashcrumb: This block also feels like it should go into the integration-support service instead of here.  Also there's already a restangular service that you can use to build this URL...

gashcrumb: There's a restangular service in the integration-support service that you could use to build this URL.

gashcrumb:
Also some lint errors:

```
src/app/integrations/list-page/list-page.component.ts[23, 19]: missing whitespace
src/app/integrations/list-page/list-page.component.ts[58, 9]: missing whitespace
```

gashcrumb:
Not sure what circle CI's deal is lately, but I wonder if the tests will run properly if you rebase against master to pick up some slight changes I made to the order of build tasks.

src/app/integrations/list-page/list-page.component.html
rhuss: @gashcrumb   Wonder whether we could introduce feature toggles in the UI, too ? So that we could switch off the feature based on some config variable. I have still the hope that we can avoid branching when we e.g. have to adapt TP2 based installations.

kahboom: That would be cool, don't see why we couldn't do this.

gashcrumb: @rhuss no reason we can't, I'd probably suggest driving that via the `config.json`

chirino: Perhaps even derive the feature toggles based on what is configured in the rest backend since type typically work hand in hand.

PR #1196, opened 2017-11-08
fix(integrations): Comment out the guided tour block for now in the integration editor

### LINKED ISSUE #1194: Can't edit an integration after creating it
This popped up this week, was definitely working last week.

If you go and try and edit an integration after creating it you either hit this kind of NPE:

screenshot 2017-11-08 14 39 48

*or* if you fix this kind of stuff locally what starts to happen is there's no error on the console, however instead of seeing the integration you want to edit you wind up in the create integration flow, i.e. the URL in the address bar doesn't show the integration ID and `edit` but instead shows `create`.

It's almost as if the router is no longer passing along the integration ID to the edit page controller, which is where we normally try and load the integration.

gashcrumb:
@kahboom I'm thinking maybe it's worth waiting to intialize the tour until we've loaded the integration, I've moved that block of code where that happens, probably also should make it conditional, as the tour probably shouldn't be started if it's a previously loaded integration, i.e. it only has an ID if we're loading a previously created integration, if it's a new integration then there's no ID.

fixes #1194

### DIFF
--- a/src/app/integrations/edit-page/edit-page.component.ts
+++ b/src/app/integrations/edit-page/edit-page.component.ts
@@ -203,7 +203,6 @@ export class IntegrationsEditPage extends ChildAwarePage
     this.routeSubscription = this.route.params
       .pluck<Params, string>('integrationId')
       .map((integrationId: string) => {
-        console.log('Integration id: ', integrationId);
         this.store.loadOrCreate(integrationId);
       })
       .subscribe();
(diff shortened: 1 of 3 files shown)


kahboom:
Most of this is outdated, sadly, if you look at my new PR. All tours are being placed in `ngOnInit`, which normally seems to suffice, as it's not always based on a single entity being loaded (I guess each case depends). Unless you disagree, I'll likely move this to `ngOnInit`. Let me know what you think though.

Not sure if I'm understanding correctly, but I don't think we should add a conditional based on whether or not an integration ID is available for that same reason, but I think once we have the steps in place we can discuss this further and improve upon it.

kahboom:
Tests failed due to a console log I think

PR #1214, opened 2017-11-10
Fix the conditional variable assignment

gashcrumb:
@chirino the `minishift-setup.sh` script didn't work for me, updated the conditional check and this seems to work fine here, can you verify?

### DIFF
--- a/scripts/minishift-setup.sh
+++ b/scripts/minishift-setup.sh
@@ -1,14 +1,11 @@
-
 #!/bin/bash
-#
-#
 
 set -euo pipefail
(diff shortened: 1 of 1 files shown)


deeleman:
I can confirm this issue. After applying the patch committed by @gashcrumb everything went back to normal, save I do not seem to be able to load the app. Although I can successfully log into the OpenShift backend app and see my projects, everytime I try to load the Syndesys project I get a `403 Permission Denied`error stating `csrf failed`.

deeleman:
Disregard my previous comment. now it works like a charm, including HMR which is a huuuuge time-saver. Thanks!!!!

PR #1216, opened 2017-11-10
fix(integrations): Bring back the trash can for the basic filter items

### LINKED ISSUE #1188: Add a way to remove rule from Basic Filter Step
There is no way to remove an existing filter step rule from Basic Filter. It causes problems when creating/editing the Basic filter step, one has the option to remove it, but I don't think this is a good way to edit the Basic filter step.
basicfilterstep

gashcrumb:
fixes #1188

### DIFF
--- a/src/app/common/ui-patternfly/syndesis-form-control.component.html
+++ b/src/app/common/ui-patternfly/syndesis-form-control.component.html
@@ -3,247 +3,250 @@
      [formGroup]="group"
      [ngClass]="[model.cls.element.container, model.cls.grid.container]">
 
-    <label *ngIf="type !== 2 && model.label !== null"
-           [attr.for]="model.id"
-           [ngClass]="[model.cls.element.label, model.cls.grid.label]">
-      <label [ngClass]="{'required-pf': model.required}">
-        {{model.label}}
-      </label>
+  <label *ngIf="type !== 2 && model.label !== null"
+         [attr.for]="model.id"
+         [ngClass]="[model.cls.element.label, model.cls.grid.label]">
+    <label [ngClass]="{'required-pf': model.required}">
+      {{model.label}}
+    </label>
     <div *ngIf="model.hint !== null"
          [tooltip]="model.hint"
          placement="right"
          class="hint-icon">
-        <span class="glyphicon glyphicon-info-sign"></span>
+      <span class="glyphicon glyphicon-info-sign"></span>
     </div>
-    </label>
+  </label>
 
-    <ng-template *ngIf="type !== 1 && template?.align === 'START'"
-                 [ngTemplateOutlet]="template.templateRef"
-                 [ngTemplateOutletContext]="model"></ng-template>
+  <ng-template *ngIf="type !== 1 && template?.align === 'START'"
+               [ngTemplateOutlet]="template.templateRef"
+               [ngTemplateOutletContext]="model"></ng-template>
 
 
-    <div *ngIf="type === 2"
-         [ngClass]="model.cls.grid.control">
+  <div *ngIf="type === 2"
+       [ngClass]="model.cls.grid.control">
 
-        <div [class.disabled]="model.disabled"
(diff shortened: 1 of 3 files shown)


src/app/common/ui-patternfly/syndesis-form-control.component.ts
```diff
@@ -48,20 +48,23 @@ export const enum SyndesisFormControlType {
...
     @Input() bindId= true;
     @Input() context: DynamicFormArrayGroupModel = null;
     @Input() group: FormGroup;
     @Input() hasErrorMessaging= false;
     @Input() model: DynamicFormControlModel;
-    @Input() nestedTemplates: QueryList<DynamicTemplateDirective>;
 
     @Output() blur: EventEmitter<DynamicFormControlEvent> = new EventEmitter<DynamicFormControlEvent>();
     @Output() change: EventEmitter<DynamicFormControlEvent> = new EventEmitter<DynamicFormControlEvent>();
     @Output() focus: EventEmitter<DynamicFormControlEvent> = new EventEmitter<DynamicFormControlEvent>();
```
deeleman: These input fields are assigned upon declaration hence do not need to be annotated with type tokens, since they will be typed by inference.

src/app/common/ui-patternfly/syndesis-form-control.component.ts
```diff
@@ -48,20 +48,23 @@ export const enum SyndesisFormControlType {
...
+    // TODO disabling this for now as the base class is in a dependency
+    /* tslint:disable */
+    @Input('templates') inputTemplates: QueryList<DynamicTemplateDirective>;
+    /* tslint:enable */
+
     @Input() asBootstrapFormGroup= true;
     @Input() bindId= true;
     @Input() context: DynamicFormArrayGroupModel = null;
     @Input() group: FormGroup;
     @Input() hasErrorMessaging= false;
```
deeleman: Spaces missing at the left of several assignation expressions here :)

src/app/common/ui-patternfly/syndesis-form-control.component.html
```diff
@@ -78,13 +78,13 @@
                                            [group]="control.at(idx)"
                                            [hasErrorMessaging]="item.hasErrorMessages"
                                            [model]="item"
-                                           [nestedTemplates]="templates"
+                                           [templates]="templates"
                                            [ngClass]="[item.cls.element.host, item.cls.grid.host]"
                                            (blur)="onFocusChange($event)"
                                            (change)="onValueChange($event)"
                                            (focus)="onFocusChange($event)"></syndesis-form-control>
```
deeleman: The problem with breaking custom element declarations into several lines per attribute like this is that code indenting becomes arbitrary and tightly coupled to the custom selector name, which is prone to change. Eg: Let's figure out we change the element tag prefix from `<syndesis-` to `<syn-` for brevity sake... Boom! All the indenting will look weird then and the template will become harder to read.

I'd suggest to stick to the default 2-spaces indenting for anything that has to be indented at some point, no matter what.

gashcrumb: That's fine, but let's do that in a separate PR.

gashcrumb: and definitely +1 2 space indenting, I just want to minimize the changes in this PR to this file to this fix, we can have a separate PR for reformatting the template files, I'd like to ensure we all have consistent formatting in our IDEs.

gashcrumb: yeah, actually this is the only file that has 4 spaces for some reason, was contributed but not sure why vscode refuses to reformat.  Sorted that via `yarn prettier`...

gashcrumb: yikes, `prettier` doesn't handle case-sensitive attributes, reverted that :-)

gashcrumb:
Okay, reformatted that template, should be all set to review again...

PR #1221, opened 2017-11-13
Reformat using Prettier, comply with TSLint

### LINKED ISSUE #1220: Reformat code base to comply with both prettier & tslint
As they are not currently aligned--if you run `yarn format`, which is quite useful for complying with style guidelines, `yarn lint` will fail due to discrepancies in the rules (e.g. semicolon). Also, much of the code base does not currently use the Prettier styling we have defined, so they need to be updated.

kahboom:
This is seemingly abusive for someone to have to review. @gashcrumb this is simply the compiled/transpiled result of running `yarn format` to comply with Prettier rules we have defined, and to ensure that it plays well with `yarn lint`.

fix #1220

### DIFF
--- a/src/app/app.component.scss
+++ b/src/app/app.component.scss
@@ -1,6 +1,8 @@
 @import 'patternfly-sass/assets/stylesheets/patternfly/color-variables';
 
-:host, .app, main {
+:host,
+.app,
+main {
   height: 100%;
 }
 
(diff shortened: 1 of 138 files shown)


kahboom:
Only way to make this less crazy is to disable from Prettier settings, but tbh I don't think there is much to review here anyway since 99.9% of it was automated (see commits).

